PR #19, opened 2015-03-04
the pointer returned by c_str only exists within the life of param

thrasibule:
Prepared statements didn't work. Currently:

```R
dbGetQuery(dbcon, "SELECT * FROM mytable WHERE var1=$1 and var2=$2", params=list("42", "18"))
```
will try to set both variables to 18.
I fixed it by copying params list into a `std::vector<std::string>`, so that the pointer doesn't get destroyed.


hadley:
Oh good catch - thanks!

PR #23, opened 2015-03-16
types_[i] can be equal to LGLSXP

thrasibule:
The function columnTypes can return LGLSXP, but this wan't picked up by columninfo which would error out.


hadley:
Thanks!

Issue #24, opened 2015-03-17
Convert dates to R native types


thrasibule:
Right now dates types are exported as character. One thing I miss from the RPostgreSQL package is that dates would be converted transparently to R types.

laurikoobas:
I'd like to see this as well. Took me a good long while to track a failing code down to a different date representation compared to RPostgreSQL. Actually I don't mind the character representation, but I sure missed it being mentioned somewhere :)

mplourde:
Additionally, `dbWriteTable` should properly convert POSIXct/Date columns to Postgres date/time types. Currently, it raises an error.

Andrei-Pozolotin:
+1

jonassundman:
This would be very nice to have implemented, as.Date conversions in R tend in my experience to be memory intensive.

PR #25, opened 2015-03-17
Convert dates

thrasibule:
This a tentative solution for #24. It probably would make more sense to convert the dates at the C++ level, but this works reasonably well for now.


hadley:
Yeah, I'd rather do this in C++, which will require a bit of work, creating a new enum for known types (instead of using SEXPTYPE) etc.

JackStat:
Is it possible to accept this pull request and then create a new issue for moving it into C++.  Dates are a real pain to deal with right now.

krlmlr:
Can you install using `devtools::install("rstats-db/RPostgres#25")` for now?

JackStat:
Yes I could but it is a year old now and a bit behind. @thrasibule can you pull and merge?

krlmlr:
Closing in favor of #111.

Issue #27, opened 2015-03-26
Include views in listed tables?


currymj:
This is also in hadley/dplyr#1007 (because it comes up when using the `tbl` function). I'm mentioning it here because it seems like RPostgres is going to replace RPostgreSQL eventually and might be able to fix the problem.

RPostgreSQL doesn't return views when using `dbListTables`, which is unexpected for some users. RPostgres uses the same query:

``` SQL
select tablename from pg_tables where schemaname !='information_schema' and schemaname !='pg_catalog';
```

I'm not sure whether RPostgres should just return views along with the tables, or whether there should be a separate `dbListViews` function. But I feel like there should be some way to get at them without writing a raw query.

The following gets both views and "real" tables:

``` SQL
SELECT table_name FROM INFORMATION_SCHEMA.tables WHERE table_schema = ANY (current_schemas(false));
```

hadley:
I think it makes sense to have this on by default.

krlmlr:
`dbListTables()` now also includes views.

PR #29, opened 2015-04-04
dbListTables now queries for view as well as table names

currymj:
Changes the query for listing tables as per #27.

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -112,8 +112,8 @@ setMethod("dbReadTable", c("PqConnection", "character"),
 #' @rdname postgres-tables
 setMethod("dbListTables", "PqConnection", function(conn) {
   dbGetQuery(conn, paste0(
-    "SELECT tablename FROM pg_tables WHERE schemaname !='information_schema'",
-    " AND schemaname !='pg_catalog'")
+    "SELECT table_name FROM INFORMATION_SCHEMA.tables ",
+    "WHERE table_schema = ANY (current_schemas(false))")
   )[[1]]
 })
 


bosr:
Hi,

I also need the ability to query views. Could you merge it please?

Thank you!
Regards

R/tables.R
```diff
@@ -112,8 +112,8 @@ setMethod("dbReadTable", c("PqConnection", "character"),
 #' @rdname postgres-tables
 setMethod("dbListTables", "PqConnection", function(conn) {
   dbGetQuery(conn, paste0(
-    "SELECT tablename FROM pg_tables WHERE schemaname !='information_schema'",
-    " AND schemaname !='pg_catalog'")
+    "SELECT table_name FROM INFORMATION_SCHEMA.tables ",
+    "WHERE table_schema = ANY (current_schemas(false))")
```
krlmlr: Does this include temporary tables?

msjgriffiths:
This query - 

```sql
SELECT table_name FROM INFORMATION_SCHEMA.tables WHERE table_schema = ANY (current_schemas(false));
```
.... only gets tables in the current `search_path`. This might be OK, but I'm not sure we should require users to have the `search_path` set correctly ahead of time. The current query, which excludes `information_schema` and `pg_catalog`, does not rely on the `search_path` being set; and pulls all schemas that are in the database (excluding the system tables). Reverting to the `search_path` might be seen as a regression.

For example, running that on a database I'm connecting to (Redshift), I get:

|   table_name      |
|-----------------------|
| table_taxonomy |

That's a real table, but it's only one of 33 real tables (it's the only table in the `public` schema, since `select current_schemas(false);` returns `{public}`).

I'd suggest instead:

```sql
SELECT table_name 
FROM INFORMATION_SCHEMA.tables 
WHERE table_schema not in ('information_schema', 'pg_catalog');
```

This produces:

|                table_name    |
|------------------------------|
| table_taxonomy|
| table1 |
| table2 |
| table3|
| ... |

(for all 33)

This should include **views**, **temporary tables**, and **tables** for all schema.

However, it might break if there are tables/views/etc with duplicate names.

krlmlr:
I'd like to postpone this until schema support is available in DBI (rstats-db/DBI#24).

krlmlr:
Thanks! Implemented a similar version in cc7a835f.

PR #41, opened 2015-07-29
add dbListFields method

### LINKED ISSUE #28: no dbListFields method for PqConnection?
I was going through your 'Basic Usage' example when I got this error message in reply to `dbListFields(con, "mtcars")`:

```
Error in (function (classes, fdef, mtable)  : 
  unable to find an inherited method for function ‘dbListFields’ for signature ‘"PqConnection", "character"’
```

Tried the same steps with a RPostgreSQL connection object with no problems. Actually:

```
> showMethods(dbListFields)
Function: dbListFields (package DBI)
conn="PostgreSQLConnection", name="character"
conn="PostgreSQLResult", name="missing"
```

whereas:

```
> showMethods(dbListTables)
Function: dbListTables (package DBI)
conn="PostgreSQLConnection"
conn="PqConnection"
```

R v3.1.3
RStudio v0.98.1103
DBI v0.3.1.9008
RPostgres v0.1

thrasibule:
fixes #28


R/tables.R
```diff
@@ -132,3 +132,13 @@ setMethod("dbRemoveTable", c("PqConnection", "character"),
...
   }
 )
+
+#' @export
+#' @rdname postgres-tables
+setMethod("dbListFields", c("PqConnection", "character"),
+  function(conn, name) {
+    name <- dbQuoteString(conn, name)
+    dbGetQuery(conn, paste("SELECT column_name FROM information_schema.columns
+WHERE table_name=", name))$column_name
```
krlmlr: This needs a bit more thought, perhaps we should add schema support: #57.

hadley: A more robust implementation might be `SELECT * FROM table_name LIMIT 0`

krlmlr: How do we detect if the failure is related to a table not found, or to some other condition? There is a `table_schema` column in `information_schema.columns`.

hadley: What other problem could it be? I think relying on these special tables is more error prone (and makes it harder to reuse code across databases).

krlmlr: Sorry: For some reason I thought this is `dbExistsTable()`. For `dbListFields()`, your approach is probably fine, although we probably still need to access internal structures for `dbExistsTable()` (which could be made faster) and `dbListTables()`: https://github.com/rstats-db/RPostgres/pull/41/files#diff-f804b88946820485d304b5ab91817450L115.

thrasibule: An argument against using "SELECT * FROM table_name LIMIT 0" is if table_name is foreign table. Foreign tables don't always push down LIMIT to the foreign backend (for instance mysql_fdw doesn't). So doing "SELECT * FROM table_name LIMIT 0" takes just as long as  "SELECT * FROM table_name", which is not acceptable if table_name is a huge table. On the other hand querying information_schema is instant.

krlmlr:
Thanks!

PR #43, opened 2015-09-18
Postgres allows a parameter called 'service' to be specified that tel…

jaredk-porch:
…ls libpq to load credentials and additional connection parameters from a central location.  This commit adds service as a parameter that RPostgres understands and will pass to libpq as an alternative to specifying parameters in the constructor.

Usage is straightforward:

```R
con <- dbConnect(RPostgres::Postgres(), service="my-service-name")
```

This will use the service called `my-service-name` defined in pg_service.conf


krlmlr:
Thanks!

PR #54, opened 2015-10-14
Added code coverage for convenience functions

JackStat:


JackStat:
One of these tests will need to be updated if/when #29 gets merged in to include views

PR #59, opened 2015-10-28
Make dbExistsTable more efficient

### LINKED ISSUE #58: Speed up dbExistsTable()
Listing all tables just for checking if one is existing is just slow: https://github.com/rstats-db/RPostgres/blob/1fcef9a543f62ffbe8e9dc1e58fc9425c871261d/R/tables.R#L123.

JackStat:
I apologize for not disentangling by pull requests.  It should be innocuous as the Oct 14th commits are supposed to boost coverage and they do test dbListTable.  Also, if there is a better way I would be happy to modify.

Fixes #58.

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -120,7 +120,11 @@ setMethod("dbListTables", "PqConnection", function(conn) {
 #' @export
 #' @rdname postgres-tables
 setMethod("dbExistsTable", "PqConnection", function(conn, name) {
-  name %in% dbListTables(conn)
+  dbGetQuery(conn, paste0(
+    "SELECT count(1) FROM INFORMATION_SCHEMA.tables ",
+    "WHERE table_schema = ANY (current_schemas(false))",
+    " AND table_name = '", name, "'")
+  )[[1]] > 0
 })
 
 #' @export


krlmlr:
I'd like to postpone this until DBI has schema support (rstats-db/DBI#24).

JackStat:
Can you provide a bit more detail. Would added schema support improve the efficiency?

krlmlr:
No, but the code needs to be different with schema support.

krlmlr:
@JackStat: Do you still think `INFORMATION_SCHEMA` is preferable over `pg_tables` here? It seems to be a part of the standard and can be expected to be more robust over time.

krlmlr:
In particular, I'm seeing leftover temporary tables when querying `pg_tables`. Not sure about `INFORMATION_SCHEMA` yet. Will investigate.

krlmlr:
Indeed, `INFORMATION_SCHEMA` doesn't seem to show these stray temporary tables:

```
$ psql -c "select * from pg_tables where tablename = 'a'" | cat
 schemaname | tablename | tableowner | tablespace | hasindexes | hasrules | hastriggers | rowsecurity 
------------+-----------+------------+------------+------------+----------+-------------+-------------
 pg_temp_2  | a         | muelleki   |            | f          | f        | f           | f
(1 row)

$ psql -c "select * from information_schema.tables where table_name = 'a'" | cat
 table_catalog | table_schema | table_name | table_type | self_referencing_column_name | reference_generation | user_defined_type_catalog | user_defined_type_schema | user_defined_type_name | is_insertable_into | is_typed | commit_action 
---------------+--------------+------------+------------+------------------------------+----------------------+---------------------------+--------------------------+------------------------+--------------------+----------+---------------
(0 rows)
```

krlmlr:
Thanks!

Issue #21, opened 2015-03-12
Fails to compile in CentOS7 (fails to find existing libpq-fe.h)


esaostring:
I try to compile "rstats-db/RPostgres", but it fails to find libpq-fe.h:

```
> devtools::install_github("rstats-db/RPostgres")
Downloading github repo rstats-db/RPostgres@master
Installing RPostgres
'/usr/lib64/R/bin/R' --vanilla CMD INSTALL  \
  '/tmp/Rtmpyy794p/devtoolsc997c6bf0ba/rstats-db-RPostgres-1f09a16'  \
  --library='/home/esa/R/x86_64-redhat-linux-gnu-library/3.1' --install-tests 

* installing *source* package ‘RPostgres’ ...
File libpq-fe.h not found. Please install libpq, e.g: postgresql-devel (rpm) or libpq-dev (deb)
ERROR: configuration failed for package ‘RPostgres’
* removing ‘/home/esa/R/x86_64-redhat-linux-gnu-library/3.1/RPostgres’
Error: Command failed (1)
```

In CentOS7, in order to get Postgresql 9.3 I use pgdg repo, and devel package includes are not in a standard place:

```
[root@localhost bin]# rpm -qil postgresql93-devel | grep libpq-fe
/usr/pgsql-9.3/include/libpq-fe.h
```

And that is also, what is reported by pg_config:

```
[root@localhost bin]# pg_config
BINDIR = /usr/pgsql-9.3/bin
DOCDIR = /usr/share/doc/pgsql
HTMLDIR = /usr/share/doc/pgsql
INCLUDEDIR = /usr/pgsql-9.3/include
[...]
```

FlorianEndel:
In CentOS 5 I had to set these environment variables prior to compiling:
```
export LIBPQ_DIR=/usr/pgsql-9.3/
export LIBRARY_PATH=/usr/pgsql-9.3/lib
```

esaostring:
Ok thanks. This seems to solve the compilation issue!

krlmlr:
Closing for now, please feel free to reopen.

dckc:
I'm using SLES 11, and I get the same symptoms.

I'm somewhat surprised that clues from `pkg-config` don't seem to be consulted.

```
> pkg-config libpq --cflags
-I/usr/include/pgsql 
```

*I don't see how to reopen this issue. I'm not sure I understand the github issue workflow.*

krlmlr:
[...]

PR #73, opened 2015-12-02
configure update for Amazon Linux

mmuurr:
Fixes the installation error of not being able locate `libpq-fe.h` for (at a minimum) Amazon Linux, which installs those header files (by default) at `/usr/include/pgsql`.
The patched version installs correctly now on Amazon Linux (4.1.7-15.23.amzn1.x86_64).


krlmlr:
Related to #21.

PR #55, opened 2015-10-16
Accept uuid types Iss30

JackStat:
If everyone is ok with uuid types being character classes in R then this is ready to merge.

Fixes #30.

PR #78, opened 2016-02-02
Return 'jsonb' as strings

jeroenooms:
The distinction between `json` and `jsonb` types only matters internally to postgres to determine if the json data gets stored in the db as an actual string or as a queryable json object. In the end, both types are the same json data and both should turn into a string in R, not a raw vector.

### DIFF
--- a/src/PqResult.h
+++ b/src/PqResult.h
@@ -272,6 +272,7 @@ private:
       case 19: // NAME
       case 25: // TEXT
       case 114: // JSON
+      case 3802: // JSONB
       case 1042: // CHAR
       case 1043: // VARCHAR
       case 1082: // DATE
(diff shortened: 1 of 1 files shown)


krlmlr:
Looks good, it was also suggested independently in #55. Do we need a test for that?

jeroenooms:
@robertzk can you comment if this is works for you?

jeroenooms:
@krlmlr seems fine to me, you can merge either one or the other...

JackStat:
I have some tests in #55 I can expand it to test that it isn't coming back in a strange way.

robertzk:
I agree that pg JSON data should be converted to a string upon reads.

krlmlr:
Thanks for the heads up. I merged #55, it came with a test, and merging this PR would most likely have introduced merge conflicts.

PR #81, opened 2016-02-24
added pg_config support; factored CFLAGS and LIBS as separate steps.

### LINKED ISSUE #22: RPostgres appears to require Homebrew installation on OS X
I attempted to give RPostgres a try, but after attempting to install all sorts of things from Homebrew, the installation gave up and suggested I uninstall MacPorts. Most of the things RPostgres was attempting to install were already present on my system.

Demand for RPostgres may increase in near future due to [uncertain status of RPostgreSQL](https://groups.google.com/forum/#!topic/rpostgresql-dev/q_aTZeO5XwY), a package that simply works for me (though compilation has become patchy with recent versions of R on OS X).

mmuurr:
The PostgreSQL installations on Amazon Linux have a variety of annoying issues.
1. They don't play nicely with pkg-config, specifically libpq doesn't have any useful .pc entries.
2. The installation location is non-standard.

But (fortunately!), pg_config does work just fine... so, the proposed changes in this pull request are:
1. Factor the PKG_CFLAGS and PKG_LIBS into two separate steps (i.e. set the values for each one separately, and check for the cascading conditions below separately for each).
2. Cascade priority as follows (from highest in the list to lowest):
- INCLUDEDIR or LIBDIR found on the command line; then
- values retrieved from pkg-config, if available; then
- values retrieved from pg_config.

This solves the problem of screwed-up pkg-config entries more generally for Linux distributions outside just Amazon Linux.
And for Amazon Linux in particular, this solves the annoying installation issues.
(I've also included a separate message for Amazon Linux users since there is no generic postgresql-devel package... instead you _must_ specify the version, e.g. postgresql94-devel.)

I'm open to suggestions for other ways to make pkg-config and pg_config play nicely together in the case where both might be installed on a single system, but the pkg-config entries be erroneous or missing (as is the case for most Amazon Linux users).

Closes #22.

### DIFF
--- a/configure
+++ b/configure
@@ -21,7 +21,7 @@ PKG_LIBS_STATIC="-lpq -lssl -lcrypto -lldap"
 # pkg-config values (if available)
 if [ $(command -v pkg-config) ]; then
   PKGCONFIG_CFLAGS=$(pkg-config --cflags --silence-errors ${PKG_CONFIG_NAME})
-  PKGCONFIG_LIBS=$(pkg-config --libs --silencer-errors ${PKG_CONFIG_NAME})
+  PKGCONFIG_LIBS=$(pkg-config --libs --silence-errors ${PKG_CONFIG_NAME})
 fi
 
 # pg_config vlaues (if available)
--- a/configure
+++ b/configure
@@ -1,36 +1,58 @@
 #!/bin/bash
-# Anticonf (tm) script by Jeroen Ooms (2016)
-# This script will query 'pkg-config' for the required cflags and ldflags.
-# If pkg-config is unavailable or does not find the library, try setting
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks, this looks good so far. I'd like to wait for feedback from other affected users before merging this.

cilindrox:
Getting the following error when testing this

```
Loading required package: methods
Error in .valueClassTest(ans, "character", "dbListTables") :
  invalid value from generic function 'dbListTables', class "NULL", expected "
character"
Calls: db_list_tables ... db_list_tables.DBIConnection -> dbListTables -> .valueClassTest
In addition: Warning message:
In value[[3L]](cond) : could not find function "columnToRownames"
Execution halted
```

Which made me think of #88

PR #82, opened 2016-03-18
Handles NA values by converting them to NULL

### LINKED ISSUE #67: Can't bind NA values
```
1. Failure: DBItest: bind_null_positional_dollar -------------------------------
is.na(rows$a) isn't true
```

DBItest: "bind_null_positional_dollar".

thrasibule:
This fixes #67.

### DIFF
--- a/src/PqResult.h
+++ b/src/PqResult.h
@@ -93,9 +93,14 @@ public:
     std::vector<int> c_formats(nparams_);
     std::vector<std::string> s_params(nparams_);
     for (int i = 0; i < nparams_; ++i) {
-      s_params[i] = Rcpp::as<std::string>(params[i][0]);
-      c_params[i] = s_params[i].c_str();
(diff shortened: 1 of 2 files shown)


src/PqResult.h
```diff
@@ -93,9 +93,14 @@ class PqResult : boost::noncopyable {
     std::vector<int> c_formats(nparams_);
     std::vector<std::string> s_params(nparams_);
     for (int i = 0; i < nparams_; ++i) {
```
krlmlr: I do wonder why only `character` parameters are supported here. Is this a limitation of PostgreSQL?

thrasibule: Everything is passed as text to the bindings. So it doesn't matter what the original type of the parameters are, they are always converted to character. We could talk to libpq using the binary protocol but that's another undertaking...

krlmlr:
Thanks!

PR #83, opened 2016-03-23
Extra connection params

### LINKED ISSUE #75: Parameter application_name isn't set on dbConnect()
Current github version of dbConnect(drv=RPostgres()) does accept application_name as a parameter. 

```
dbConnect(drv=RPostgres::Postgres(), application_name = 'applicationX')
```

However when executing
`SELECT * FROM pg_stat_activity`

The application_name has not been set.

I am currently doing the following as a work around

``` r
con <- dbConnect(drv=RPostgres::Postgres())
           dbSendQuery(con, "SET application_name = 'applicationX'")
           dbDisconnect(con)
```

Check
`SELECT * FROM pg_stat_activity`
again and you can see the application_name has been set

thrasibule:
This closes #75. I had to export the pid in dbGetInfo to facilitate testing.

### DIFF
--- a/R/connection.R
+++ b/R/connection.R
@@ -62,7 +62,7 @@ setMethod("dbConnect", "PqDriver", function(drv, dbname = NULL,
   host = NULL, port = NULL, password = NULL, user = NULL, service = NULL, ...) {
 
   opts <- unlist(list(dbname = dbname, user = user, password = password,
-    host = host, port = as.character(port), service = service, client_encoding = "utf8"))
+    host = host, port = as.character(port), service = service, client_encoding = "utf8", ...))
   if (!is.character(opts)) {
     stop("All options should be strings", call. = FALSE)
   }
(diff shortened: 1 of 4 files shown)


R/connection.R
```diff
@@ -62,7 +62,7 @@ setMethod("dbConnect", "PqDriver", function(drv, dbname = NULL,
   host = NULL, port = NULL, password = NULL, user = NULL, service = NULL, ...) {
 
   opts <- unlist(list(dbname = dbname, user = user, password = password,
-    host = host, port = as.character(port), service = service, client_encoding = "utf8"))
+    host = host, port = as.character(port), service = service, client_encoding = "utf8", ...))
```
krlmlr: Are there no restriction on the naming of options?

thrasibule: The options name are arbitrary, they're all listed here: http://www.postgresql.org/docs/9.4/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS. We can add them all as keywords arguments, or pass them in the .... The way the code was written, the ... wasn't used at all, even though it was documented as a way to pass extra options. If you pass an incorrect option, it throws an error with an informative error message, so I think it's fine not filtering the options explicitely.

krlmlr:
Thanks!

Issue #50, opened 2015-10-04
Warning with untyped NULL


krlmlr:
`SELECT NULL as a` leads to the following warning:

In result_create(conn@ptr, statement) :
  Unknown field type (705) in column a

Related: #30

Test name: "data_null".

krlmlr:
The same error occurs when selecting string constants.

krlmlr:
Types are defined here: http://doxygen.postgresql.org/include_2catalog_2pg__type_8h_source.html

Should be used in PqResult.h instead of hard-coded numbers. Also think about mapping of NULL (2278, 705).

hadley:
Hard coded numbers are easier because the type catalog changes with each version.

PR #89, opened 2016-08-25
Handle string quoting and missing values in strings

jimhester:
I ran into this when trying to use `dbWriteTable(copy = FALSE)`. While the column and row names were escaped, the data values were not and `connection_escape_string` did not handle missing values.

Behavior after this PR

``` r
con <- dbConnect(RPostgres::Postgres(), db = "test_db", host = "localhost", password = "password", port = "5432")
dbQuoteString(con, c("AS", NA))
#> <SQL> 'AS'
#> <SQL> NULL
```

Missing values in non-character atomic vectors are still an issue, they don't seem to be handled currently.

### DIFF
--- a/R/quote.R
+++ b/R/quote.R
@@ -19,7 +19,10 @@ NULL
 #' @export
 #' @rdname quote
 setMethod("dbQuoteString", c("PqConnection", "character"), function(conn, x, ...) {
-  SQL(connection_escape_string(conn@ptr, x))
+  is_na <- is.na(x)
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks. I'm getting DBItest failures when applying this to current master, could you please double-check? Also, I was hoping that this would help with #50, but the relevant tests still fail (as you mentioned).

krlmlr:
Thanks!

Issue #91, opened 2016-08-30
RPostgres-specific strInterpolate, sqlAppendTableTemplate


r2evans:
RPostgres accepts parameters with `$1`, `$2`. The function from DBI assumes `?1`, `?2` format.
### Expected Behavior:

When used, the functions should provide dbengine-specific for parameter indexing. That is, for RSQLite, one of `?`, `?1`, or `:name`; for RPostgres, `$1`; for RSQLServer `?` (limited by jTDS). Because I don't expect `DBI` to track all other engines' mechanisms, I would expect this method to be overridden by each driver.
### Actual Behavior:

``` r
connpg <- dbConnect(RPostgres::Postgres())
dbWriteTable(connpg, "mtcars", mtcars[1,], row.names = FALSE)
# [1] TRUE
(qry <- sqlAppendTableTemplate(connpg, "mtcars", mtcars[,3:5], row.names = FALSE))
# <SQL> INSERT INTO "mtcars"
#   ("disp", "hp", "drat")
# VALUES
#   (?1, ?2, ?3)
res <- dbSendQuery(connpg, qry)
# Error: ERROR:  operator does not exist: ? integer
# LINE 4:   (?1, ?2, ?3)
#            ^
# HINT:  No operator matches the given name and argument type(s). You might need to add explicit type casts.
(qry2 <- gsub("\\?", "$", qry))
# <SQL> INSERT INTO "mtcars"
#   ("disp", "hp", "drat")
# VALUES
#   ($1, $2, $3)
res <- dbSendQuery(connpg, qry2)
dbBind(res, mtcars[3:5])
dbClearResult(res)
# [1] TRUE
```

(The insertion is actually incomplete, just inserting the first row, but I'm guessing that's a separate issue ... likely my mistake.)

r2evans:
Disregard, fixed with `prefix`:

```r
(qry <- sqlAppendTableTemplate(connpg, "mtcars", mtcars[,3:5], prefix = "$", row.names = FALSE))
# <SQL> INSERT INTO "mtcars"
#   ("disp", "hp", "drat")
# VALUES
#   ($1, $2, $3)
```

PR #97, opened 2016-09-11
add sqlAppendTable method for PqConnection

r2evans:
related to (but still different from) #91, allows:

``` r
dbExecute(con, sqlAppendTable(con, tblname, dat))
```

(Used in a scenario where I need to get auto-populated via `"INSERT INTO table (...) VALUES (...) RETURNING id"`.)


r2evans:
For the appveyor build, I don't understand why this error is there:

> `Project file "RPostgres.Rproj" is not a valid MSBuild project file which must be an XML.`

I didn't modify the file, and though it is certainly not xml, it is present in both branches. (I'll fix the travis error in a bit. I also have a couple of other tweaks I didn't catch before submitting.)

krlmlr:
Thanks. I don't understand why we need to override `sqlData()` here -- `DBI::sqlData()` already seems to be generic enough. What's the behavior of RPostgres without this PR?

krlmlr:
I got confused. So you did actually override `sqlAppendTable()`, but why is *that* necessary?

r2evans:
I've gone back to my use-cases, and it must have been something else that was pushing me to suggest the change.

PR #108, opened 2016-12-17
Fix roundtrip logical

### LINKED ISSUE #60: dbWriteTable() with logical column fails
```
1. Error: DBItest: roundtrip_logical -------------------------------------------
ERROR:  invalid input syntax for integer: "true"
CONTEXT:  COPY test, line 1, column a: "true"
```

DBItest: "roundtrip_logical".

### LINKED ISSUE #107: Can't write table with a logical column
I would expect this to successfully import

```r
library(DBI)
con <- dbConnect(RPostgres::Postgres())
dbWriteTable(con, "tmp", data.frame(a = TRUE, b = 1))
```

but I get an error from the server:

```
Error: ERROR:  invalid input syntax for integer: "true"
CONTEXT:  COPY tmp, line 1, column a: "true"
```

With RSQLite, this converts to integer.

If I have an existing table with boolean columns this works fine - the import does the conversion to TRUE/FALSE values (and `dbReadTable` converts back to logical when reading back into R).

This is with the current master of RPostgres (d694bd446bdd8f2a99b77594096f07c3c183cd1b, version 0.1-2) and with postgres running in a docker container (PostgreSQL 9.6.0 on x86_64-pc-linux-gnu, compiled by gcc (Debian 4.9.2-10) 4.9.2, 64-bit)

thrasibule:
This closes #60 and closes #107

### DIFF
--- a/R/connection.R
+++ b/R/connection.R
@@ -105,7 +105,7 @@ get_data_type <- function(obj) {
     integer = "INTEGER",
     double = "REAL",
     character = "TEXT",
-    logical = "INTEGER",
+    logical = "BOOLEAN",
     list = "BLOB",
     stop("Unsupported type", call. = FALSE)
   )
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks!

PR #111, opened 2016-12-20
DateTime parsing

thrasibule:
This supersedes PR #25, and does the date conversions in C++ directly. This still might not be the most efficient way to do things, since apparently mktime is pretty slow as it is doing system calls to figure out what the local timezone is.
I represent times as hms, so I use a forked version of DBItest which handles this see thrasibule/DBItest@5950ad4. Using this branch, tests pass locally for me.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -19,7 +19,8 @@ Depends:
 Imports:
     methods,
     Rcpp (>= 0.11.4.2),
-    DBI (>= 0.3.1.9008)
+    DBI (>= 0.3.1.9008),
+    hms
 Suggests:
     testthat,
     DBItest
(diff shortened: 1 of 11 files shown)


krlmlr:
Thanks. I'll need to review the DBItest changes separately before looking at this PR. Would you mind submitting them? (Need to use namespace imports instead of library(), I'd also like to make use of the hms package optional. Coercion to/from a `difftime` object should be sufficient.)

krlmlr:
Thanks again. Are you still interested in this pull request? I'd like to merge the refactorings separately (e.g., using `PGTypes` instead of `SEXPTYPE`).

thrasibule:
Yes, having native date and datetime support would be really useful. Do you want me to send a seperate pull request with just the PGTypes change first?

krlmlr:
Yes please, this would be great

krlmlr:
Thanks!

PR #112, opened 2017-01-09
Handle bigint

parisni:
Hey,

I had two surprises today:

1. bigint were not handle well by RPostgres (both pull & push). This lead to invisible error, and was difficult to debug eg: 8998329829898429834 -> 1878743324
2. When trying to modify the code, I learned that R do not handle int8. The usual solution is to cast them as character (https://www.r-bloggers.com/r-in-a-64-bit-world/)

Then find here the patch that handle postgresql bigint as strings. 
BTW, dbWriteTable will create TEXT field.

My guess is this is inevitable.

### DIFF
--- a/src/PqResult.h
+++ b/src/PqResult.h
@@ -254,7 +254,6 @@ private:
       Oid type = PQftype(pSpec_, i);
       // SELECT oid, typname FROM pg_type WHERE typtype = 'b'
       switch(type) {
-      case 20: // BIGINT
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks. We'll offer better support for 64-bit integers in the future, leaving this open as a reference.

krlmlr:
bigint handled now via `bigint` argument to `dbConnect()`.

PR #149, opened 2017-12-03
Fix build for MacOS

jeroen:
...


krlmlr:
Thanks. In what situations is this important?

jeroen:
Mostly for building the OSX binary packages (once you submit to cran).

Issue #114, opened 2017-01-23
Round trip geometry columns


hadley:
Using simple features

etiennebr:
I'd like to give a hand, however I'm not sure how to do this properly, I'd appreciate if you could please advise me. 

I've thought of solutions relying on `dbWriteTable`. 

1. I believe we could update `get_data_type` to create the table with `geometry` type, but it uses `typeof`, which returns `list` for every custom object. Also I'm not sure that detecting `sfc` type belongs to RPostgres, so maybe you have another approach to suggest. This would ultimately enable to write `sfc` using `dbWriteTable`. I read discussion on DBI issues about handling custom formats, so there might be a solution coming?

2. Another option would be to expose the `value` argument in `sqlCreateTable` so we can properly override the default `BYTEA` for lists without needing to tweak `get_data_type`. Then we could use a custom `sf` writer such as current `st_read_db`. Something like:
```r
st_get_data_type <- function(obj) {
  if(inherits(obj, "sfc")) {
  	gtype <- st_geometry_type(obj)[1]
  	crs <- st_crs(obj)
  	d <- st_dimension(nc$geometry)[1]
    return(geom = paste0("geometry(", gtype, ",", crs[["epsg"]],",", d, ")"))
  }
  return(RPostgres:::get_data_type(obj))
}
#' This is a proposal, real function would require additional check of geometry type and dimension uniformity
```
If we could then pass a `vapply(sf, st_get_data_type, "")` to  `dbWriteTable`, then I believe we could create the table properly. Also, in favor of this approach is that prior to Posgis 2.0.0, the `geometry_columns` table had to be edited (e.g. using `AddGeometryColumn` which would be impossible using `dbWriteTable` directly. Though I think we can focus on post 2.0.0 as it will make it much simpler, but maybe something to consider.

3. Last option is to duplicate code form `dbWriteTable` in a `st_write_db`, then we can add whatever operation we need, but I believe `RPostgres` is meant to be pretty generic, so my understanding is that there is a way to make it work. Also other major issue is that it requires to rewrite the `st_write_db` for every driver, which would defy the purpose of a DBI, but might also be the pragmatic way. 

For reading the geometry, I'm not clear if type casting relies on `Rcpp::wrap`? If so, then I guess we'll have to rely on the current procedure: read > cast to text using dbRead > find geometry columns and crs from database > cast geometries to `sfc`.

edzer:
[...]

PR #173, opened 2018-02-14
Typname

etiennebr:
@krlmlr I'm opening this P/R so you can have some visibility. Let me know how it looks so far. 

A few highlights to help you navigate changes:
* Exposes a hook named `finalizeType`, a generic S3 method, to make it easy for other packages to add types. 
* The `pg_types` are cached in the `PqConnection` object.
* I surfaced the oids using an attribute to the ` DbDataFrame`, maybe you will see something better, (I wasn't sure about my choice)
* There are probably existing functions I could use to `finalizeTypes`, `typeLookup`,  `append` and `yankClass`.

#114, r-dbi/DBI#203

### DIFF
--- a/R/PqResult.R
+++ b/R/PqResult.R
@@ -44,7 +44,7 @@ setMethod("dbGetRowsAffected", "PqResult", function(res, ...) {
 #' @export
 setMethod("dbColumnInfo", "PqResult", function(res, ...) {
   rci <- result_column_info(res@ptr)
-  typeLookup(rci, res@conn)
+  cbind(rci, typname = typeLookup(rci[["oid"]], res@conn), stringsAsFactors = FALSE)
 })
 
 #' Execute a SQL statement on a database connection
@@ -138,22 +138,28 @@ finalizeTypes <- function(ret, conn) {
 
   typnames <- vapply(attr(ret, "oids"), typeLookup, character(1), conn)
   types <- vapply(ret, class, character(1))
-  class_edit <- types %in% "character" & !typnames %in% "unknown"
+  class_edit <- types %in% "character"
   ret[class_edit] <- mapply(appendClass, ret[class_edit], typnames[class_edit], SIMPLIFY = FALSE)
   ret[] <- lapply(ret, finalizeType)
   structure(ret, oids = NULL)
 }
 
+set_class <- function(x, subclass = NULL) {
+  class(x) <- c(subclass)
+  x
+}
+
 appendClass <- function(x, .class) {
-  structure(x, class = c(.class, class(x)))
+  set_class(x, c(.class, class(x)))
 }
 
 yankClass <- function(x) {
   if(length(class(x)) < 2) return(x)
-  structure(x, class = class(x)[-1])
+  set_class(x, class(x)[-1])
 }
 
 typeLookup <- function(x, conn) {
+  if(length(x) > 1) return(vapply(x, typeLookup, character(1), conn))
   with(conn@typnames, typname[oid == x])
 }
 
--- a/R/PqResult.R
+++ b/R/PqResult.R
@@ -43,7 +43,8 @@ setMethod("dbGetRowsAffected", "PqResult", function(res, ...) {
 #' @rdname PqResult-class
 #' @export
 setMethod("dbColumnInfo", "PqResult", function(res, ...) {
-  result_column_info(res@ptr)
+  rci <- result_column_info(res@ptr)
+  merge(res@conn@typnames, rci, by = "oid")
 })
 
 #' Execute a SQL statement on a database connection
--- a/R/PqResult.R
+++ b/R/PqResult.R
@@ -44,7 +44,7 @@ setMethod("dbGetRowsAffected", "PqResult", function(res, ...) {
 #' @export
 setMethod("dbColumnInfo", "PqResult", function(res, ...) {
   rci <- result_column_info(res@ptr)
-  merge(res@conn@typnames, rci, by = "oid")
+  typeLookup(rci, res@conn)
 })
 
 #' Execute a SQL statement on a database connection
@@ -117,7 +117,8 @@ setMethod("dbFetch", "PqResult", function(res, n = -1, ..., row.names = FALSE) {
   if (is.infinite(n)) n <- -1
   if (trunc(n) != n) stopc("n must be a whole number")
   ret <- sqlColumnToRownames(result_fetch(res@ptr, n = n), row.names)
-  convert_bigint(ret, res@bigint)
+  ret <- convert_bigint(ret, res@bigint)
+  finalizeTypes(ret, res@conn)
 })
 
 convert_bigint <- function(ret, bigint) {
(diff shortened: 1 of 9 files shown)


hadley:
Can you outline the basic idea? `appendClass()` and `yankClass()` are ringing warning bells to me — it's usually bad practice to manually the class vector in this way.

src/PqResultImpl.cpp
```diff
@@ -134,13 +135,25 @@ std::vector<DATA_TYPE> PqResultImpl::_cache::get_column_types(PGresult* spec)  {
 
     default:
```
hadley: Shouldn't all the lookup here use the cached type table?

etiennebr: Yes, I'm not convinced either. I'd say it's a shortcut. When I discussed it with @krlmlr he wasn't fond of changing the logic for all the types, so I opted for this option.

The casting is performed as before, we just remove the warning. Special types are passed as characters  and then the `finalizeType` hook is exposed to any package that wants to finalize a vector of type `character`.

Here are the alternatives I considered, (maybe there's a better option I haven't considered):
* Cast all types to `characters` in `get_column_type` then use a `finalize` in R to cast to the correct type. (sounds inefficient)
* `finalize` in cpp, but then I don't know how to provide a hook for other packages. It would be an R call for generic functions from cpp. (This is probably the best solution, if it's possible)
* Cast known types, others to `character` and then use a `finalize` for these. (That's the proposed solution.)

Also consider, and maybe that was your question, that some types are described in `pg_type.h` while other are defined by the extensions. Hence the different handling. See https://stackoverflow.com/a/12531823/171659 for details.

R/PqResult.R
```diff
@@ -131,6 +133,55 @@ convert_bigint <- function(ret, bigint) {
   ret
 }
 
+finalizeTypes <- function(ret, conn) {
+  if(nrow(conn@typnames) < 1) return(ret)
+
+  typnames <- vapply(attr(ret, "oids"), typeLookup, character(1), conn)
+  types <- vapply(ret, class, character(1))
+  class_edit <- types %in% "character"
```
etiennebr: @hadley, also see here that the typname is only added to `character` columns.

etiennebr:
The strategy indeed ~~abuses~~ leverages the S3 flexibility since it adds the `typname` from pg to the class to allow other functions to access it. I'm very open to other strategies. 

Thanks for your feedback.

src/PqResultImpl.cpp
hadley: Ah ok, makes sense.

hadley:
I'd like to see a sketch of the S3 interface, and some thinking about how to avoid clashes with other S3 classes. I'd think maybe base class of "pq_type", and then subclasses would have `pq_` + the postgres type name.  We'd have a R level helper for testing, but in regular usage the class would be set in C++ (which will require some replumbing of the existing data structures) - that I believe would eliminate the need for `finalizeTypes`.

I don't think `finalizeType.default()` actually needs to do anything - it should return `x` unchanged, which would eliminate the need for `yankClass()`.  

`finalizeType()` needs a different name:

* I think it should use snake case to make it clear that it's not a S4 generic (because dynamic S4 classes are v. hard, which also needs to be explained in the docs.
* I think it's best to avoid names with different UK/US spelling
* I think something that hints that it's about coercing to an R type would be better.

etiennebr:
@hadley I'm not sure what you mean by *a sketch of the S3 interface*. This is the best (somewhat working) prototype I could come up with, so far, and I feel it enables the conversation, but if you think another form of documentation could help, I'll be happy to provide it! 

✔ I agree that classes should be prefixed to avoid collisions. I'd go for `class = c("pq_type", "pq_<typname>")` where `<typname>` is the returned typname (`json` or `geometry` for instance). That's also what @krlmlr suggested (he used `pg`, but this will likely create collisions).

Doing this, we can mostly get rid of `yankClass`. Now the question is: should the class (`"pq_type", "pq_<typname>"`) be carried around after coercion or should all unknown types be cast to characters by `coerce_pg_type` (to replace `finalizeType`)? I like your suggestion to have minimal intervention and pass these classes, I just don't know exactly how it will behave ⚠. For instance I was surprised to see that `SELECT pg_typeof('a');` is `unknown`. I could try to find most of these corner cases and provide specific `coerce_pg_type` methods (`coerce_pg_type.pq_unknown`, `coerce_pg_type.pq_bpchar`) directly in RPostgres, though.

✔Setting the classes in c++ might have a positive impact for very wide tables, I guess. I was thinking to use a std::map for the lookup. 

I would also hide the `typname` table from R (in this prototype available as a slot in `PqConnection`).

In the very unlikely event of a missing `oid`, should it `stop()` or pass the `oid` in replacement of the `typname` like `c("pq_type", "pq_3491")`. It doesn't mean anything, but rather than destroy the query, it allows the user to handle the type manually.

hadley:
I think at better way to attack this PR would have been to sketch out the basic approach in prose and then convert it to code. This would involve describing the S3 interface (e.g. all non-hardcoded types will get a custom class ...) then implementing the R side (e.g. `new_pg_type()`), then implementing the C++ side.

I don't see any downside to keeping the classes around. There shouldn't be any harm associated with keeping them, and then that information is available if the user needs it. I also don't see the need to hide the `typename` table from R.

etiennebr:
Ok, I apologize for ruching this PR with no introduction. I felt like the discussion happened in the issues mentioned above, and the few discussions with @krlmlr at RConf. I felt it was clear, but now I realize it was probably only to both of us. Thanks for taking the time to figure it out @hadley.

Let me give it another shot. Here's the broad plan. The problem is that: 
1. some blobs retrieved from the database are of type unknown to R, and
2. RPostgres raises a warning when it receives these types

Point 2. is trivial, we can just remove the warning (which will remove the need to hard-code *known* Oids in `get_column_types()`). As for the special types (1.), most common types are hard-coded in PostgreSQL in `catalog/pg_types.h` some of these most common types are already handled in `RPostgres`; they include `char`, `float`, `bool`, ... To resolve the other blob types there is a `pg_type` table that describes all types available in the database. These types can be created by the user or more likely by an extension, such as `postgis` or simply not handled natively by R, such as `json`. 

Special types are currently coerced to `character` column during the transition from the database to R `data.frame`. A `character` column in R can then be coerced to a specific class, but it requires prior knowledge of which column contains which class, or additional query to the database. Since the information on the column type is available when fetching results, but it is currently not returned with the `data.frame`, making it hard to retrieve which column should be coerced to which type. 

In this PR, I want to a) implement a way to attach the original database type to the column and b) provide a function for packages to coerce known types using S3 generic function. Note that a) can be implemented without b). 

My initial intention was to keep the returned type as close as it currently is. I mean by that to only affect special types, and when a special-type-column cannot be coerced, return it as a character, just as it used to be, so it is transparent for the user). Essentially keeping the special-types usage internally (but opened to other packages). After our exchanges, I believe it might be acceptable to plainly expose special types and thus change some expected results (e.g. by `DBItest`). 

Special types will receive the `pq_type` class (since `pg_type` is already a type name in PostgreSQL) and a second custom class `pq_` + `<typname>`. This will provide continuity from the database typename to R, which was not available before.

Since I didn't want to expose the special-types classes (to affect the returned type as little as possible), my plan was to provide access to the special-types-column right before returning the resulting `data.frame`.  Access would be provided by a S3 generic `coerce_pg_types` function. After our exchanges, I'm not sure this is mandatory anymore. We could just stop there and provide these custom classes for any function that wants to digest it. It would also make results more consistent since the output wouldn't vary depending on the packages loaded.

The current PR mostly demonstrates the ability to resolve oids to typnames and attach them to the returned `data.frame`. It also provides a generic function (`finalizeType`, to be renamed `coerced_pg_type`, but maybe reduced in scope) to enable access to known types for other packages. It caches the `pg_type` table on connect and uses it to assign classes to the `data.frame`'s columns (currently in R, will be moved to c++). 

Hopefully this clarifies the intention. You mentioned testing, I'm not accustomed to testing C++ from R, so I appreciate suggestions or pointers to best practices or good examples.

R/PqResult.R
```diff
@@ -131,6 +133,55 @@ convert_bigint <- function(ret, bigint) {
...
+  set_class(x, c(.class, class(x)))
+}
+
+yankClass <- function(x) {
+  if(length(class(x)) < 2) return(x)
+  set_class(x, class(x)[-1])
+}
+
+typeLookup <- function(x, conn) {
+  if(length(x) > 1) return(vapply(x, typeLookup, character(1), conn))
```
krlmlr: Can we use `match()` here?

R/PqResult.R
```diff
@@ -131,6 +133,55 @@ convert_bigint <- function(ret, bigint) {
   ret
 }
 
+finalizeTypes <- function(ret, conn) {
+  if(nrow(conn@typnames) < 1) return(ret)
+
+  typnames <- vapply(attr(ret, "oids"), typeLookup, character(1), conn)
```
krlmlr: `typeLookup()` seems to be vectorized already.

src/DbDataFrame.cpp
```diff
@@ -57,9 +58,17 @@ List DbDataFrame::get_data(std::vector<DATA_TYPE>& types_) {
   out.attr("names") = names;
   out.attr("class") = "data.frame";
   out.attr("row.names") = IntegerVector::create(NA_INTEGER, -i);
+  out.attr("oids") = as<IntegerVector>(wrap(oids));
   return out;
 }
 
+std::vector<Oid> get_oids() {
+  std::vector<Oid> oids;
+  oids.push_back(Oid(991));
```
krlmlr: Why are these values special?

krlmlr:
@etiennebr: Updated review comment.

etiennebr:
Thanks for your feedback @krlmlr and @hadley; I'll try to submit changes this week. 

I didn't know that `DbDataFrame` was identical across db's; I'll try to re-route OID's to `PqDataFrame` using the `Pq*` classes. 

I'm using `typname` because it's the name used by PostgreSQL, maybe to avoid conflicts with C's reserved `typename`, but we could indeed rename it to `typename` or simply `type`?

krlmlr:
Thanks, Etienne. We can keep `typname` if it's adapted from Postgres code, maybe explain this decision somewhere in a comment?

krlmlr:
Thanks! I'll tweak myself. I think it's too early to implement `finalizeType()`, but the type information will be available in the `"class"` attribute.

PR #193, opened 2018-08-25
Add user interrupt check while waiting for query results to be ready.

zozlak:
Sometimes user wants to terminate a query before a first row of results is available (e.g. it turned out that the query is too complicated and he wants to rewrite it instead of waiting for its completion). So far user interrupts where not checked until `PQgetResult()` completion so the only option to achieve that was to terminate whole R session or to terminate the query on the database server.

This pull requests adds user interrupts check between `PQsendQueryPrepared()` and `PQgetResult()`.

It is done by waiting for data on the database connection socket with a timeout (1 second) using glibc's `select()` and running `checkUserInterrupt()` if the timeout is hit.


src/PqResultImpl.cpp
```diff
@@ -379,6 +381,35 @@ bool PqResultImpl::step_run() {
   LOG_VERBOSE;
 
   if (pRes_) PQclear(pRes_);
+
+  // Check user interrupts while waiting for the data to be ready
```
krlmlr: Could you please extract this to a method?

zozlak: Extracted in f164a77

DESCRIPTION
```diff
@@ -1,11 +1,12 @@
...
 Title: 'Rcpp' Interface to 'PostgreSQL'
-Version: 1.1.1
-Date: 2018-05-05
+Version: 1.1.2
+Date: 2018-08-25
 Authors@R: c(
     person("Hadley", "Wickham", role = "aut"),
     person("Jeroen", "Ooms", role = "aut"),
     person("Kirill", "M\u00fcller", role = c("aut", "cre"), email = "krlmlr+r@mailbox.org", comment = c(ORCID = "0000-0002-1416-3412")),
+    person("Mateusz", "\u017b\u00f3\u0142tak", role = c("ctb")),
```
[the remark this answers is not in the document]
zozlak: Moved to NEWS.md in d6dcba9

zozlak:
> The new code contains some libpq function calls I'm not familiar with, could we make this optional for now?

Of course we can. How would you prefer it to be done?

By the way I added a link to the PG documentation in the code and an example of using `PQconsumeInput()` can be found at https://www.postgresql.org/docs/current/static/libpq-example.html (they don't use `PQisBusy()` in the example because they are waiting for notifications and not data but the idea is the same)

krlmlr:
Thanks. I think an argument to `dbConnect()` would be a good way.

Can you please also resolve the conflicts? I can take care of the NEWS entry.

zozlak:
Conflicts removed, adding `dbConnect()` parameter requires a little more time :)

zozlak:
Any idea how to name the parameter?

zozlak:
I added it as `checkInterrupts` but I'm not attached to the name :)

src/PqResultImpl.cpp
```diff
@@ -453,3 +463,34 @@ void PqResultImpl::add_oids(List& data) const {
 PGresult* PqResultImpl::get_result() {
   return pRes_;
 }
+
+// checks user interrupts while waiting for the first row of data to be ready
+// see https://www.postgresql.org/docs/current/static/libpq-async.html
+void PqResultImpl::wait_for_data() {
+  int socket, ret;
+  fd_set input;
+  timeval timeout = {0, 0};
```
krlmlr: Move definition of variable into loop, because `select()` can overwrite the contents.

krlmlr:
Could you please double-check if the code still works as intended after my tweaks? In `master` now.

zozlak:
Seems to work fine :)

Just a very minor remark - as you moved the `timeval timeout` declaration inside the loop in the `src/PqResultImpl.cpp:485` it can be written as 
```
timeval timeout = {1, 0};
```
instead of
```
timeval timeout = {0, 0};
timeout.tv_sec = 1;
```

PR #204, opened 2018-12-12
change test from REAL to DOUBLE PRECISION

harvey131:
Resolve issue in test case caused by _Translate floating-point values to `DOUBLE PRECISION`_ on Nov 26 in this commit:  https://github.com/r-dbi/RPostgres/commit/67449d5d64ef36ad200f4abb5226663a31735d2a


krlmlr:
Thanks!

PR #207, opened 2019-01-02
Windows: update to libpq 11.1.0

jeroen:


krlmlr:
What a leap. Do we need to expect breaking changes?

jeroen:
I don't think so, homebrew has also upgraded without any issues.

krlmlr:
Thanks!

Issue #187, opened 2018-06-06
Enhancement: Allow for disabling `SET TIMEZONE`


brascoball:
I know this isn't the proper use for this package, but RPostgres works perfectly for a [JBoss Data Virtualization Layer](https://developers.redhat.com/products/datavirt/overview/) on top of an [Amazon Redshift](https://aws.amazon.com/redshift/) data warehouse.  My team has been using it for over a year now. The only issue is that I had to fork the repo and comment out:
`dbExecute(con, "SET TIMEZONE='UTC'")` 

in **PqConnection.R** because that's not a valid command in JBoss DV. Is there any way you could add a parameter for RPostgres() to allow for not running that command?

Thanks for considering!

krlmlr:
Thanks, I think we should support a `timezone` argument for `dbConnect()` across all backends, with default `"UTC"` and `NULL` meaning that the time zone isn't set. Would you like to contribute a PR?

comicfans:
+1 for this feature . RPostgres use UTC which is different to psql (local timezone) , which cause lots of confusion. I'd like DBI to have ability to specified timezone for this scenario.

trafficonese:
The PR should close that issue.

Anyway, I am still seeing different results to `RPostgreSQL`. 

With RPostgreSQL I could just do, to get the expected result:
`select time_stamp from table;`

With RPostgres I have to write the query like this to get the same results:
`select time_stamp at time zone 'Europe/Vienna' from table;`

The time_stamp column is of type `timestamp with time zone`, the DB is on Ubuntu and I am running on Windows.

Can someone enlighten me, why that is the case? Is it because of the different DB-Driver?

krlmlr:
Closed by #214.

@trafficonese: Can you please open a new issue, with a reproducible example that shows the problem? Ideally the example creates the table it needs to show the behavior you're seeing.

PR #214, opened 2019-03-12
add timezone argument to dbConnect

trafficonese:
Default is UTC, NULL/NA omits setting the timezone.

Adresses #187


krlmlr:
Thanks! I'll tweak a bit.

PR #220, opened 2019-05-28
dbUnquoteIdentifier

baileych:
Modifies RPostgres' own dbUnquoteIdentifier() to parse catalog.schema.table specifications without each component being enclosed by double-quotes, as required to work with dbplyr.

Fixes r-dbi/DBI#277.

### DIFF
--- a/R/quote.R
+++ b/R/quote.R
@@ -71,21 +71,25 @@ setMethod("dbQuoteIdentifier", c("PqConnection", "Id"), function(conn, x, ...) {
 #' @export
 #' @rdname quote
 setMethod("dbUnquoteIdentifier", c("PqConnection", "SQL"), function(conn, x, ...) {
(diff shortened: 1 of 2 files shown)


baileych:
The update breaks an existing test which passes to dbQuoteIdentifier an `Id` object with a schema name and no table.  The resulting SQL string has the form `"schema_name".`, which does not conform to the dbQuoteIdentifier() spec's requirement that the result be suitable as a table or column name.  The new dbUnquoteIdentifier() rejects this as an illegal SQL expression (correctly), but this does mean that a dbQuoteIdentifier()/dbUnquoteIdentifier() round trip for an arbitrary `Id` object is not guaranteed.

I would argue that the new behavior is correct, given the spec for dbQuoteIdentifier().  If so, the fix is to change the list_objects_features test in spec-sql-list-objects.R to attempt quoting/unquoting only objects that have a table name (i.e. is_prefix is false).

If the preference is to retain the roundtrip rather than a valid SQL expression, the fix is to change the regex in dbUnquoteIdentifier() to accept zero or more characters after a `.` rather than 1 or more.

PR #230, opened 2019-11-27
Hide unused symbols in shared library

troels:
Hi, 

I have problems using RPostgres and RSQLite from the same R-process.

I get stacktraces from R such as these:

```
 *** caught segfault ***
address 0x1000000010, cause 'memory not mapped'

Traceback:
 1: result_release(res@ptr)
 2: dbClearResult(rs)
 3: dbClearResult(rs)
 4: dbExecute(con, "SET TIMEZONE='UTC'")
 5: dbExecute(con, "SET TIMEZONE='UTC'")
 6: .local(drv, ...)
 7: DBI::dbConnect(RPostgres::Postgres(), dbname = "kingfisher")
 8: DBI::dbConnect(RPostgres::Postgres(), dbname = "kingfisher")
 9: eval(expr)
10: eval(expr)
11: create_connection(production = production)
12: load_list_of_tables("ensembl", tables, production = production)
An irrecoverable exception occurred. R is aborting now ...
bin/load_ensembl.sh: line 7: 36907 Segmentation fault      (core dumped) Rscript R/load_ensembl.R
```

After wading through the code with gdb it appears that RPostgres and RSQLite export a number of symbols named the same, which are then randomly assigned to the first loaded of the libraries, leading to crashes, stack corruption and other nastiness.

I therefore suggest changing default symbol visibility to hidden, so the two libraries can coexist peacefully. A number of other mysterious issues, *might* also be consequences of this symbol conflict.

I created a corresponding pull request for RSQlite: r-dbi/RSQLite#303

See here for more information:
https://github.com/RcppCore/Rcpp/pull/720


.travis.yml
```diff
@@ -58,6 +58,7 @@ addons:
     - libpq-dev
     - libmagick++-dev
     - libssh2-1-dev
+    - devscripts
```
krlmlr: I wonder what this is for.

krlmlr:
Thanks for the PRs. Does this problem also affect {RMariaDB} and {odbc}?

.travis.yml
troels: Hi @krlmlr 

It's for the script checkbashisms, which is required by "R CMD check" in R-devel now.

troels:
Yeah, odbc and RMariaDB would benefit from similar patches too. I might be able to get to it in the next few days, if I can find a little free time.

Issue #229, opened 2019-11-24
Incorrect results when aggregating timestamp to date


cswingle:
There seems to be a problem with RPostgres when retrieving results involving aggregating `timestamp with time zone` columns to date. The SQL that is generated (displayed with `show_query()`) yields a different result than what is actually retrieved by RPostgres. RPostgreSQL generates the same SQL, and retrieves the correct result.

In the following example, focus on the 2019-05-01 data, which should have a minimum of 24 and a maximum of 47.

```{r}
library(tidyverse)
library(lubridate)
library(RPostgres)

conn <- dbConnect(Postgres(),
                  host = "dbhost",
                  user = "dbuser",
                  dbname = "dbname")

example <- tibble(val = seq(0:70)) %>%
    mutate(ts = ymd_hms("2019-04-30 00:00:00", tz = "US/Alaska") + hours(val))

example %>%
    mutate(dte = date(ts)) %>%
    group_by(dte) %>%
    summarize(min = min(val, na.rm = TRUE),
              max = max(val, na.rm = TRUE))
```

That yields the correct result:

```
  dte          min   max
  <date>     <int> <int>
1 2019-04-30     1    23
2 2019-05-01    24    47
3 2019-05-02    48    71
```

If we round-trip this through the database we get a different answer:

```{r}
dbWriteTable(conn, "example", example,
             overwrite = TRUE, append = FALSE)

conn %>%
    tbl("example") %>%
    mutate(dte = date(ts)) %>%
    group_by(dte) %>%
    summarize(min = min(val, na.rm = TRUE),
              max = max(val, na.rm = TRUE))
```

Here's the result. The values for min and max appear to be off by 8 hours as though it's treating the timestamp of the data in the database as UTC (rather than US/Alaska).

```
  dte          min   max
  <date>     <int> <int>
1 2019-04-30     1    15
2 2019-05-02    40    63
3 2019-05-03    64    71
4 2019-05-01    16    39
```

It's not the query that is generated, which gives the correct result when run inside Postgres. I get the same query when I load RPostgreSQL and use it's database driver, but it loads the proper results.

Here's the `show_query()` output:

```
<SQL>
SELECT "dte", MIN("val") AS "min", MAX("val") AS "max"
FROM (SELECT "val", "ts", date("ts") AS "dte"
FROM "example") "dbplyr_057"
GROUP BY "dte"
```

And the results of that query when run inside the database using `psql`:

```
    dte     | min | max
------------+-----+-----
[...]

PR #231, opened 2019-11-27
WIP: Set the default timezone to NULL on dbConnect

cswingle:
This resolves Issue #229 (when built with DBI >= 1.0.0.9003)


trafficonese:
I think you missed to update the documentation.
Thats why travis is nagging.

cswingle:
@trafficonese Thanks. You're right. I updated the help page to match the internal documentation.

cswingle:
Not sure what to do about the Travis CI build fail. It looks to me like maybe it's this: `A complete check needs the 'checkbashisms' script`, but I haven't a clue how to fix that. When I run `devtools::check(error_on = "warning")` on my system I get a clean check (Linux, AMD64, R 3.6.1, DBI and friends from github).

trafficonese:
I was wondering about that aswell, then I saw the commit https://github.com/r-dbi/RPostgres/pull/230/commits/27c079860c25b7c52ff75ce01a239c72a683da5a of #230 which apparently fixed the `checkbashisms` problem.

cswingle:
Thanks. I just saw that too and did the same thing with 5c9a349. Fingers crossed.

cswingle:
I've been playing with this pull request, and while it does solve the issue with #229 it breaks normal retrieval of `timestamp with time zone` data from Postgres. With this patch in place, it doesn't seem possible to retrieve the correct timestamp value regardless of what value you use in the `timestamp` argument to `dbConnect`. Back to the drawing board on #229, I guess.

krlmlr:
Thanks. Let me keep it open as a WIP to see what hasn't worked.

krlmlr:
Thanks again. This should be now fixed with #276. I suspect we need to keep the default for `timezone` to `"UTC"`.

PR #256, opened 2020-09-08
fix_posixt: format as ISO-8601

### LINKED ISSUE #255: fix_posixt() formats timestamps to ambiguous timezone
fix_posixt() calls format_keep_na(col, usetz = T) which in terms calls base::format(...) to format POSIXt columns. However, the formatted time could contain ambiguous timezone and result in wrong time when parsed by PostgreSQL.

For example:

2020-09-01 12:00:00 Asia/Shanghai is formatted to: "2020-09-01 12:00:00 CST", which will be parsed by PostgreSQL as Central Standard Time (North America), a 14-hour difference.

Possible solution:

I suggest calling format(t, format = "%Y-%m-%dT%H:%M:%OS%z") to format timestamps to ISO-8061 to avoid such problems.

```r
t <- lubridate::as_datetime("2020-09-01 12:00:00", tz = "Asia/Shanghai")
# "2020-09-01 12:00:00 CST" ambiguous! will be parsed as Central Standard Time by PostgreSQL
format(t, usetz = TRUE)
# "2020-09-01T12:00:00+0800" not ambiguous, will be correctly parsed
format(t, format = "%Y-%m-%dT%H:%M:%OS%z")
```

imlijunda:
Fixes #255

### DIFF
--- a/R/PqResult.R
+++ b/R/PqResult.R
@@ -195,7 +195,7 @@ fix_posixt <- function(value) {
   is_posixt <- vlapply(value, function(c) inherits(c, "POSIXt"))
   withr::with_options(
     list(digits.secs = 6),
-    value[is_posixt] <- lapply(value[is_posixt], function(col) format_keep_na(col, usetz = T))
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks!

Issue #251, opened 2020-08-17
materialized views in dbListTables()


dpprdan:
Should Materialized Views be listed by `dbListTables()`? 

I'd argue they should since they are objects that share characteristics of both Tables and Views, both of which are returned by `dbListTables()`, see #27 & #29. 

See [this thread](https://www.postgresql.org/message-id/flat/CAH7T-ao6ece1mgCsCvsE04W59ZZJP9gGXVK97wkUzRV5gsDqQQ%40mail.gmail.com) on the psql-hackers list for a discussion on why they are not in the `INFORMATION_SCHEMA.tables` (which includes Views). Tl;dr: "They are not defined by the SQL standard." I see this as a relatively minor point w.r.t. RPostgres, though. 

As it stands there is no helper function like `dbListTables()` to list Materialized Views in RPostgres. 

The following query shows all Tables, Views and Materialized Views (I have not looked into temporary tables, yet): 

```sql
SELECT tablename AS name FROM pg_tables WHERE schemaname = ANY (current_schemas(false))
UNION
SELECT viewname AS name FROM pg_views WHERE schemaname = ANY (current_schemas(false))
UNION
SELECT matviewname as name FROM pg_matviews WHERE schemaname = ANY (current_schemas(false))
ORDER BY name;
```

dpprdan:
Probably a better implementation and one that is closer to the current one:

```sql
SELECT c.relname AS name
FROM   pg_class AS c 
  JOIN pg_namespace AS n 
  ON   c.relnamespace = n.oid
WHERE  (n.nspname = ANY (current_schemas(true)))
  AND  (n.nspname <> 'pg_catalog')
  AND  (relkind IN ('r', 'p', 'f', 'v', 'm'))
  AND NOT relispartition
ORDER BY name
```

This also includes foreign and partitioned tables like `INFORMATION_SCHEMA.tables`. 

`AND NOT relispartition` omits the partitions of a parrtitioned table. Tables with many partitions would otherwise clutter the list and I don't see a use-case for accessing the partitions directly from R. I might be missing something, though.

krlmlr:
Thanks. This sounds like a useful extension. Would you like to submit a pull request?

dpprdan:
I submitted a PR for `dbListTables()` (#261), but now realise that this probably should extend to `dbExistsTable()` and `dbListFields()` as well?! 

[...]

Issue #215, opened 2019-04-04
Better support for RedShift


krlmlr:
- #164 
- #188 
- #211 
- #212 

Perhaps with a separate subclass.

Existing solutions:

- https://cran.r-project.org/web/packages/RRedshiftSQL/index.html
- https://github.com/pingles/redshift-r
- https://github.com/sicarul/redshiftTools

kmishra9:
Ran into #211 as well

isteves:
@krlmlr I currently use Redshift in case I can be of help with testing/etc.

kmishra9:
A teammate of mine pointed out coming across this thread, so I thought I'd leave the functional replacement to `copy_to` that I wrote up until it's implemented in the Driver packages! It's super suboptimal, relying on the very old redshiftTools package, but better than nothing, in a pinch!

```
#*******************************************************************************
# Organization - Cricket Health
# Description - Functional replacement for dplyr::copy_to(), which doesn't work w/ Redshift
#*******************************************************************************

upload_df_to_rs <-
    function(df,
             rs,
             rs_schema = 'public',
             rs_relation_name = deparse(substitute(df)),
             overwrite = TRUE) {
        #' @Description: A function that works similarly to dplyr::copy_to() that uploads a local dataframe to Redshift via temporarily saved files in S3
        #' @param df: the target dataframe to upload
        #' @param rs: a database connection object (the result of a DBI::dbConnect()) call, indicating the database where the data should be uploaded
        #' @param rs_schema: a string, indicating the schema within rs to copy the file to; example: "lnd"
        #' @param rs_relation_name: a string, indicating the table within rs_schema to copy the file to
        #' @param overwrite: a boolean, indicating whether an existing table at the path {rs}.{rs_schema}.{rs_relation_name} should be dropped first or not. If one exists and overwrite is not TRUE, an error will occur
        #' @Returns: A dbplyr table reference to the uploaded df

        require('tidyverse')
        require('glue')

        table_path <-
            glue::glue('{rs_schema}.{rs_relation_name}')

        table_paths <-
            rs %>%
            tbl(in_schema('information_schema', 'tables')) %>%
            collect() %>%
            transmute(table_path = glue('{table_schema}.{table_name}')) %>%
            distinct() %>%
            pull(table_path)

        rs_schemas <-
            rs %>%
            tbl('pg_namespace') %>%
            select(table_schema = nspname) %>%
            distinct() %>%
            pull(table_schema)

[...]

Issue #211, opened 2019-02-21
dbExistsTable fails on Amazon Redshift


ilyaminati:
`dbExistsTable` on Redshift returns:

> Error in result_create(conn@ptr, statement) : 
> Failed to prepare query: ERROR:  function generate_subscripts(name[], integer) does not exist
HINT:  No function matches the given name and argument types. You may need to add explicit type casts.

This is likely because `generate_subscripts` is [unsupported on Redshift](http://awsdocs.s3.amazonaws.com/redshift/latest/redshift-dg.pdf).

krlmlr:
Thanks. I'm not familiar with RedShift, can you suggest an alternative query?

https://github.com/r-dbi/RPostgres/blob/e09b0147cbe4bc4c6fb3bf806cab1ee556dc52f9/R/tables.R#L269-L276

ilyaminati:
Sorry, I no longer have access to a Redshift instance, I'm no help here. But what's the reason for doing such a search in the first place? For `dbExistsTable`, is it not sufficient to just query against [PG_TABLE_DEF](https://docs.aws.amazon.com/redshift/latest/dg/r_PG_TABLE_DEF.html)?

krlmlr:
Sure, but will this work for Postgres?

krlmlr:
We don't currently support Redshift very well. Closing in favor of #215.

PR #261, opened 2020-09-28
dbListTables: list Materialized Views as well

dpprdan:
`dbListTables()` now uses the [`pg_class`](https://www.postgresql.org/docs/current/catalog-pg-class.html)/[`pg_namespace`](https://www.postgresql.org/docs/current/catalog-pg-namespace.html) tables instead of `INFORMATION_SCHEMA.tables` to retrieve the names of remote tables and (materialized) views. 

In particular `dbListTables()` will now list:

- r = ordinary table
- v = view
- m = materialized view
- f = foreign table
- p = partitioned table

It does not list the partitions of a partitioned table.


dpprdan:
There are no `dbListTables()` tests, correct?

krlmlr:
I'm getting legitimate Travis errors: https://travis-ci.com/github/r-dbi/RPostgres/jobs/391997882#L3128

dpprdan:
☹️ I can reproduce on a Postgres 9.6. Apparently, `relispartition` (as well as `p = partitioned table`) was added in v10 (pg_class docs [v10](https://www.postgresql.org/docs/10/catalog-pg-class.html) vs [v9.6](https://www.postgresql.org/docs/9.6/catalog-pg-class.html)). 

How do you want me to handle this? 

1. Drop `relispartition` i.e. list partitions as well.
2. Make the query dependent on the postgres version, i.e. drop relispartition only for Postgres < v10.

Relatedly: Do you already have a policy regarding the use of [information schema](https://www.postgresql.org/docs/current/information-schema.html) vs. [system catalogs](https://www.postgresql.org/docs/9.6/catalogs.html)?

> The information schema consists of a set of views that contain information about the objects defined in the current database. The information schema is defined in the SQL standard and can therefore be expected to be portable and __remain stable — unlike the system catalogs__, which are specific to PostgreSQL and are modeled after implementation concerns. The information schema views do not, however, contain information about PostgreSQL-specific features; to __inquire about those you need to query the system catalogs__ or other PostgreSQL-specific views.
https://www.postgresql.org/docs/current/information-schema.html (emphasis mine).

Since mviews are PostgreSQL-specific features we need to query the system catalogs. These have the drawback of being less stable, which we see here to some degree. 

Therefore, alternative no.

3. Do not support PostgreSQL-specific features and therefore mviews in `dbListTables()` (and [`dbListFields()` and `dbExistsTable()`](https://github.com/r-dbi/RPostgres/issues/251#issuecomment-700009224)).

BTW Which Postgres versions does RPostgres support?

krlmlr:
A robust solution would be better. We require client libraries >= 9.6, the server version is not specified.

Are materialized views contained in `INFORMATION_SCHEMA.views` ?

dpprdan:
> Are materialized views contained in INFORMATION_SCHEMA.views ?

No, they are not since they are a PostgreSQL-specific feature. For more info, see [this thread](https://www.postgresql.org/message-id/flat/CAH7T-ao6ece1mgCsCvsE04W59ZZJP9gGXVK97wkUzRV5gsDqQQ%40mail.gmail.com) on the psql-hackers list for a discussion on why they are not in the INFORMATION_SCHEMA.tables (which includes Views). Tl;dr: "They are not defined by the SQL standard." 

My 2ct: Since RPostgres is a PostgreSQL-specific package, it makes sense to support PostgreSQL-specific features. 

OTOH: Since it is a DBI package, maybe it might also make sense to only support SQL-standard features (like information schema)? 

In this particular case I don't expect major changes in the future, but you never know. 

> We require client libraries >= 9.6, the server version is not specified.

Apparently my mental model is off. What do you mean by client libraries? I thought it goes like this: RPostgres - libpq - PostgreSQL server.

krlmlr:
We require libpq >= 9.6 .

I'm also worried about Redshift (#215). Can we implement a solution that works on all platforms?

dpprdan:
I am not familiar with Redshift, at all, and don't know how to test ATM.

[DESCRIPTION](SystemRequirements: libpq >= 9.0: libpq-dev (deb) or) says `libpq >= 9.0`, BTW.

dpprdan:
I found this:

> The standard PostgreSQL catalog tables are accessible to Amazon Redshift users. For more information about PostgreSQL system catalogs, see [PostgreSQL system tables](https://www.postgresql.org/docs/8.0/static/catalogs.html#CATALOGS-OVERVIEW).

This links to the Postgres v8.0 docs, however. Like I said, I am not familiar with Redshift, so I don't know whether this is an oversight or not.

I think we have a decision though: In order to support system catalogs, which are less stable than information schema, we would have to test against all currently supported server versions. If you want to support Redshift as well, we would need to test against all current Redshift versions as well. This is probably not as easy as spinning up a few docker containers (though I don't know for sure). This sounds very much like "let's not support queries against system catalogs in RPostgres" to me. 

What do you think?

krlmlr:
Right, 9.0 is stated in DESCRIPTION and in `./configure` .

I don't mind implementing Postgres-specific code now that we have `Redshift()` as a driver for which we can special-case. I could set up a Redshift test database to check this, also on CI.

krlmlr:
We could find a middle ground by implementing Postgres-specific functions to gain some experience: Leave `db*Table*()` unchanged and implement Pq-specific functions with a distinct prefix.

dpprdan:
Sounds good, I think! Just to make sure, you mean:

| INFORMATION_SCHEMA | systems catalog |
|-----------------|-----------------|
| `dbListTable()`   | `pqListTables()`  |
| `dbListFields()`  | `pqListFields()`  |
| `dbExistsTable()` | `pqExistsTable()` |
| `dbListObjects()` | `pqListObjects()` |

This involves a bit more than just changing a SQL query, though. E.g. I will have to read up on S4. Also we may have to think about how we setup internal functions like `find_table()`. Or rather I will have to figure out how this is designed elsewhere in the package, at least.

What I am saying is: I won't be able to do this immediately and cannot commit to a time frame at this moment. Also, I may require a bit of support from you. Therefore it's also fine with me, if you want to do this yourself. 😃

dpprdan:
Probably obvious, but we may need Redshift-specific functions as well, when a query relies on a Postgres function that [Redshift does not support](https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-functions.html), e.g. https://github.com/r-dbi/RPostgres/issues/211. (Plus unsupported PG [data types](https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-datatypes.html) and [features](https://docs.aws.amazon.com/redshift/latest/dg/c_unsupported-postgresql-features.html)).

Not the best example, though, because this particular case seems to have been fixed in the meantime, since `dbListTables()` does not use the `generate_subscripts` function anymore. 

https://github.com/r-dbi/RPostgres/blob/12c3d3dc67f93cbafca24b5496d2fc1a9cf0f562/R/tables.R#L237-L244

dpprdan:
I added new pq* versions of `dbListTables()`, `dbExistsTable()`, `dbListObjects()` and `dbListFields()`. They all query the system catalog (pg_class in particular) and are based on the same [underlying code](https://github.com/r-dbi/RPostgres/pull/261/files#diff-73a35922acced633ebf3fbf9b9a7c59f3b9ee52ddd3b0b67f4781216cf5850c1R46-R99). Remember that we have to use the system catalog instead of information schema because materialized views are not included in information schema (because they are not in the SQL spec). 

Is this, what you had in mind, @krlmlr?

At the minimum this needs some more tests to make sure that it does what we are trying to do, i.e. list materialized views as well. 

> We could find a middle ground by implementing Postgres-specific functions to gain some experience: Leave db*Table*() unchanged and implement Pq-specific functions with a distinct prefix.

In the end I think it would be better to use this new code for the db* functions instead of having these "special" functions. One reason is ease of use, another is downstream functions (e.g. in [dplyr](https://github.com/tidyverse/dplyr/issues/1007)), which build on the db* functions. 
We could still use information schema based code for connections that do not support (this) pg_class code, i.e. for Redshift connections.

Slightly off-topic: Do you have plans to [test non-current PostgreSQL](https://github.com/r-dbi/RPostgres/blob/d04e86307ee20abd39b3f2a53c31f70f3378057e/.github/workflows/R-CMD-check.yaml#L146) versions with R CMD check on GHA? I am asking because it was set up on Travis and [caught an error](https://github.com/r-dbi/RPostgres/pull/261#issuecomment-699987760) I made.

krlmlr:
Thanks a lot. I'm not sure we need the new `pq*()` functions to be S4 methods, also this package uses the `postgres` prefix in similar situations.

Other than that, I think we can expose this via the `db*()` interface if the user sets a flag in `dbConnect()` -- similarly to the `check_interrupts` flag. Tests would be great too.

Added checks for Postgres >= 10 to the build matrix.

dpprdan:
Closing this in favour of #414

PR #264, opened 2020-10-27
Fix dbQuoteLiteral() for difftime values

krlmlr:
.


krlmlr:
Will use interval instead.

PR #282, opened 2021-01-08
Fix typo

jeroen:
I think this causes https://www.r-project.org/nosvn/R.check/r-oldrel-macos-x86_64/RPostgres-00install.html


krlmlr:
Thanks!

PR #283, opened 2021-01-10
Fix configure

jeroen:
In https://github.com/r-dbi/RPostgres/commit/620d011ee57aa70050a2f4b666bbb1cfd6259bba you seem to have copied some partial changes, but we need this part as well because later down we `cat configure.log` in case of failure.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -17,3 +17,4 @@ windows
 ^LICENSE\.md$
 ^CODE_OF_CONDUCT\.md$
 ^tests/testthat/testthat-problems\.rds$
+^.*\.log$
--- a/configure
+++ b/configure
@@ -86,10 +86,10 @@ echo "Using PKG_CFLAGS=$PKG_CFLAGS"
 echo "Using PKG_LIBS=$PKG_LIBS"
 
 # Test configuration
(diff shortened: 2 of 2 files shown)


krlmlr:
Thanks. I've received more feedback:

> More issues:
>
> 1) The M1mac additional issue shows a typo in the configure script.
>
> 2) I have never seen pkg-config nor a .pc file natively on macOS, so the comments need evidence or to be removed.
>
> 3) A design error of many hand-written configure scripts is that there is no way to ask them to use 'autobrew' or internal code, or to not use that.  As soon as there are other packages using the same external software (and there is, 'rpg') we need to have that installed and then it is used in preference to autobrew etc, which is likely what the maintainer tested.  I have spent many hours on this for gert/git2r and asked the git2r maintainer for a configure option to unjam the conflicts -- please also provide one for RPostgres.

I think we fixed the typo in `configure`, what do we do for the two other issues?

jeroen:
For (2) I suppose we just remove the comment.

For (3) this is interesting, he seems to want a way to force using autobrew. I have to think about that a bit.

krlmlr:
@jeroen: Did you get a chance to look at the "force-autobrew" thing? I need to release very soon.

jeroen:
OK I think it should work now.

krlmlr:
Thanks!

PR #284, opened 2021-01-15
Fix check_tz to return desired result when timezone is invalid

ateucher:
A small thing I noticed:

When an invalid timezone is passed to the `timezone_out` argument in `dbConnect()`, the correct warning is thrown but the invalid timezone is kept due to a bug in `check_tz` (should be `""`):

``` r
library(RPostgres)

con <- postgresDefault(timezone = "+01:00")
#> Warning: Invalid time zone '+01:00', falling back to local time.
#> Set the `timezone` argument to a valid time zone.
#> CCTZ: Unrecognized output timezone: "+01:00"
con@timezone
#> [1] "+01:00"
dbDisconnect(con)

RPostgres:::check_tz("+01:00")
#> Warning: Invalid time zone '+01:00', falling back to local time.
#> Set the `"+01:00"` argument to a valid time zone.
#> CCTZ: Unrecognized output timezone: "+01:00"
#> [1] "+01:00"
```

<sup>Created on 2021-01-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This PR fixes this by assigning`timezone` to the parent env with `<<-` within `tryCatch` in `check_tz`, and adds a couple of new tests.

### DIFF
--- a/R/PqConnection.R
+++ b/R/PqConnection.R
@@ -233,7 +233,11 @@ check_tz <- function(timezone) {
   arg_name <- deparse(substitute(timezone))
 
   tryCatch(
-    lubridate::force_tz(as.POSIXct("2021-03-01 10:40"), timezone),
+    {
+      # Side effect: check if time zone is valid
+      lubridate::force_tz(as.POSIXct("2021-03-01 10:40"), timezone)
+      timezone
+    },
     error = function(e) {
       warning(
         "Invalid time zone '", timezone, "', ",
@@ -242,11 +246,9 @@ check_tz <- function(timezone) {
         conditionMessage(e),
         call. = FALSE
       )
-      timezone <<- ""
+      ""
     }
   )
-
-  timezone
 }
 
 # dbDisconnect() (after dbConnect() to maintain order in documentation)
--- a/R/PqConnection.R
+++ b/R/PqConnection.R
@@ -242,7 +242,7 @@ check_tz <- function(timezone) {
         conditionMessage(e),
         call. = FALSE
       )
-      timezone <- ""
+      timezone <<- ""
     }
   )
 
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks, good catch. I think the intention was to have `check_tz()` return the repaired time zone, no need to assign to the parent environment. Would you like to tweak?

ateucher:
Thanks @krlmlr. As submitted, `check_tz()` does return the repaired time zone... it's the error handler in`tryCatch()` that assigns the repaired timezone to the `check_tz()` environment (which is then returned).

Unless I'm overthinking it, the other way to do it is to wrap `lubridate::force_tz` in `lubrdate::tz()`, and assign `tryCatch()` to the `timezone` object (see below):

``` r
check_tz <- function(timezone) {
  arg_name <- deparse(substitute(timezone))
  
  timezone <- tryCatch(
    lubridate::tz(lubridate::force_tz(as.POSIXct("2021-03-01 10:40"), timezone)),
    error = function(e) {
      warning(
        "Invalid time zone '", timezone, "', ",
        "falling back to local time.\n",
        "Set the `", arg_name, "` argument to a valid time zone.\n",
        conditionMessage(e),
        call. = FALSE
      )
      ""
    }
  )
  timezone
}

check_tz("Etc/GMT+8")
#> [1] "Etc/GMT+8"
check_tz("not_a_timezone")
#> Warning: Invalid time zone 'not_a_timezone', falling back to local time.
#> Set the `"not_a_timezone"` argument to a valid time zone.
#> CCTZ: Unrecognized output timezone: "not_a_timezone"
#> [1] ""
```

<sup>Created on 2021-01-18 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

krlmlr:
Thanks, I was too confused by `<<-`; changed it to a pattern similar to yours.

ateucher:
Oh that is better - thanks!

PR #292, opened 2021-02-12
Force autobrew libs on CRAN

### LINKED ISSUE #291: Recent update causes error on sslmode
I reverted my RPostgres back to 1.2.1 because recent version causes error related to ssl.
Here's error I got :

```r
con = dbConnect(Postgres(),
                  host='wrds-pgdata.wharton.upenn.edu',
                  port=9737,
                  dbname='wrds',
                  sslmode='require',
                  user='Matthew')

#sslmode value "require" invalid when SSL support is not compiled in
```

Before rolling back, I tried installing (updating ) openssl and libpq, but didn't help.

jeroen:
Fixes #291.

This is what happened: CRAN recently [updated libpq on their server](https://github.com/R-macos/recipes/commit/2a186b954cf95b0169779cdbd9ad19f71c879da3) and now it exposes the `.pc` file. Therefore our package was picking up on that libpq, instead of the autobrew build. However the CRAN build is very old and does not support ssl.

### DIFF
--- a/configure
+++ b/configure
@@ -16,6 +16,11 @@ PKG_BREW_NAME="libpq"
 PKG_TEST_HEADER="<libpq-fe.h>"
 PKG_LIBS="-lpq"
 
+# Workaround for old libpq without ssl on CRAN servers
+if [ -d "/Volumes/Builds/packages" ] || [ -d "/Volumes/SSD-Data/Builds" ]; then
+FORCE_AUTOBREW=1
+fi
+
 # pkg-config values (if available)
 if [ -z "$FORCE_AUTOBREW" ] && [ `command -v pkg-config` ]; then
   PKGCONFIG_CFLAGS=`pkg-config --cflags --silence-errors ${PKG_CONFIG_NAME}`


krlmlr:
Thanks. Postgres 9.6 still seems to be supported, it's not entirely unreasonable to build against an older library. I suspect this hack will be noticed, either now or at any point in the future. The directory structure can also change.

Can we instead use Homebrew by default, and make use of other libraries opt-in? Or perhaps check if the configured libpq supports SSL (how?), and force autobrew on if unsupported, with a message?

I'm told that

> I have never seen pkg-config nor a .pc file natively on macOS, so the comments need evidence or to be removed.

Does libpq 9.6.20 count as evidence here?

jeroen:
I don't think there is an easy way to see if a local libpq supports ssl or not.

We could make autobrew the default on macos if you like, perhaps that makes sense in this case. Usually I would think that one would prefer a locally available system library; but in this case the local one seems almost always bad.

PR #294, opened 2021-03-25
Extra checks for Macos

jeroen:
Fixes https://github.com/r-dbi/RPostgres/issues/291 Closes https://github.com/r-dbi/RPostgres/pull/292

Another solution, specifically on MacOS, to prevent linking against outdated or non-ssl versions of libpq.


krlmlr:
Thanks!

PR #297, opened 2021-04-04
run scheduled workflows only in r-dbi repo

dpprdan:
The scheduled workflows currently run in forked repos as well. 
E.g. 144 workflow runs of the 158 in [my fork](https://github.com/dpprdan/RPostgres/actions) are scheduled runs.

grafik

Am I right to assume that this is neither necessary nor intended?

I therefore configured the workflows in such a way that they only run on a schedule in repositories owned by https://github.com/r-dbi. (Alternatively one could [condition on the repository](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#github-context), i.e. `r-dbi/RPostgres`, but conditioning on the owner makes it easier to transfer to other r-dbi repos.)

### DIFF
--- a/.github/workflows/R-CMD-check-dev.yaml
+++ b/.github/workflows/R-CMD-check-dev.yaml
@@ -11,6 +11,7 @@ name: rcc dev
 
 jobs:
   matrix:
+    if: (github.event_name != 'schedule') || (github.event_name == 'schedule' && github.repository_owner == 'r-dbi')
     runs-on: ubuntu-18.04
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -22,6 +22,7 @@ name: rcc
 
 jobs:
   R-CMD-check:
+    if: (github.event_name != 'schedule') || (github.event_name == 'schedule' && github.repository_owner == 'r-dbi')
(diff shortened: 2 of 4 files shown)


dpprdan:
The pkgdown workflow generally only makes sense in `r-dbi` or on pull requests, I suppose?

krlmlr:
Thanks. I see the problem of wasted compute resources.

I'd rather not hard-code the repository owner in the workflow files, because these workflows are used for different owners and should be as portable as possible.

Can you disable Actions on your fork?

Are workflows enabled for new forks? I just forked two new-to-me repos, via UI and via API, and see this:

Screenshot from 2021-04-05 05-06-08

Can you confirm? Pick any of https://krlmlr.github.io/actions-sync/ if you need a new fork.

Also, scheduled workflows will cease running after 60 days.

Relevant: https://news.ycombinator.com/item?id=24655458.

dpprdan:
> Can you disable Actions on your fork?

You're right, this is probably the easiest solution. I had thought about that and was worried that this would disable the `push` and `pull_request` events as well (which would be neat to have as a check, especially when developing for different platforms/versions). But the push events are only relevant for the main branch (on which I shouldn't be developing) and the PR events still work on, well, PRs to the main repo. 

> Are workflows enabled for new forks? 

Apparently they are not. I do not remember enabling them for the RPostgres fork, though, so maybe this is a recent change to GHA, i.e. they do not only disable scheduled but all workflows? I dunno.

Issue #244, opened 2020-04-21
check_interrupts = TRUE causes certain query with inner join to fail with error message "select() on the connection failed"


LaugeGregers:
The following query, when run through `dbGetQuery()` fails when using `check_interupts = TRUE` in `dbConnect()`. Error message is "select() on the connection failed"

```SQL
select dsc.dw_customer_key
from dws_customer dsc
inner join tmp_20200420_customer_reg_dates_2019 tmp 
  on tmp.dw_customer_key = dsc.dw_customer_key
where 1 = 1 and dsc.dw_distribution_key between 201902 * 1337 and 201903 * 1337
limit 10
```

The queries are run against my company's Data warehouse. 

Queries including only one of the tables `dws_customer` or `tmp_20200420_customer_reg_dates_2019` work. Queries with other inner joins work fine.

```r
bad_query <- "select dsc.dw_customer_key
from dws_customer dsc
inner join tmp_20200420_customer_reg_dates_2019 tmp 
  on tmp.dw_customer_key = dsc.dw_customer_key
where 1 = 1 and dsc.dw_distribution_key between 201902 * 1337 and 201903 * 1337
limit 10"

# Shows that we can fetch from dws_customer
working_query <- "select dsc.dw_customer_key 
from dws_customer dsc 
limit 10"

# Shows that we can fetch from tmp_20200420_customer_reg_dates_2019
working_query2 <- "select tmp.dw_customer_key
from tmp_20200420_customer_reg_dates_2019 tmp
limit 10"

# shows that we can do inner joins 
working_query3 <- "select 
  dfc.dw_contact_event_key, 
  ddc.external_info1_description
from dwf_contact dfc
inner join dwd_contact_event ddc 
    on ddc.dw_contact_event_key = dfc.dw_contact_event_key
limit 10"

library(DBI)

# Following fails on last query
tmp_con <- 
  dbConnect(RPostgres::Postgres(),
            dbname = "dw_db",
            host = "company DW", 
            port = "proper port", 
            user = "foobar", 
            password = "foobar",
            check_interrupts = TRUE)
dbGetQuery(tmp_con, working_query)
dbGetQuery(tmp_con, working_query2)
dbGetQuery(tmp_con, working_query3)
dbGetQuery(tmp_con, bad_query) # Error: select() on the connection failed
dbDisconnect(tmp_con)
```

Changing the order in which queries are run, doesn't change the result. 

If we set `check_interrupts = FALSE` the query works. The others work as well.

[...]

PR #303, opened 2021-04-17
Issue 244

zozlak:
As reported in #244 the `check_interrupts` connection option didn't work for Windows user when a query execution took longer than 1 second.

This pull requests fixes the issue by adjusting the code to the Windows implementation of the `select()` glibc/winsock method:

* The `fd_set` structure is reinitialized with the `FD_SET` macro before each call to `select()`
* `select()` error message now includes the error code and the error code is captured in the platform-specific way (`WSAGetLastError` on Windows, `errno` global variable on other platforms)
* A unit test is added.

### DIFF
--- a/tests/testthat/test-checkInterrupts.R
+++ b/tests/testthat/test-checkInterrupts.R
@@ -6,5 +6,4 @@ test_that("Query is executed properly when check_interrupts connection option is
   con <- postgresDefault(check_interrupts = TRUE)
   expect_equal(dbGetQuery(con, "SELECT pg_sleep(2), 'foo' AS x")$x, "foo")
   dbDisconnect(con)
-  expect_null(n)
 })
--- a/tests/testthat/test-checkInterrupts.R
+++ b/tests/testthat/test-checkInterrupts.R
@@ -1,0 +1,10 @@
+context("checkInterrupts")
+
+test_that("Query is executed properly when check_interrupts connection option is 
+          set to TRUE and query execution takes more than check user interrupts 
+          timeout (1 second)", {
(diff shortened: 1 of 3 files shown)


tests/testthat/test-checkInterrupts.R
```diff
@@ -0,0 +1,10 @@
+context("checkInterrupts")
+
+test_that("Query is executed properly when check_interrupts connection option is 
+          set to TRUE and query execution takes more than check user interrupts 
+          timeout (1 second)", {
```
krlmlr, line 5:
```suggestion
test_that("check_interrupts = TRUE works with queries > 1 second (#204)", {
```

krlmlr (commented):
Thanks for looking into it. I've had a great experience with https://github.com/mxschmitt/action-tmate/ to SSH into a worker on GitHub Actions.

tests/testthat/test-checkInterrupts.R
zozlak: Sure, you just made a typo in the issue number. I committed the shortened description with the right number already.

zozlak:
The ubuntu-18.04 (devel) build seems to fail on not my code :-)

zozlak:
@krlmlr how should we proceed? The devel@ubuntu18.04 fail doesn't seem to be connected with my code.

krlmlr:
Thanks. I'll merge the next time I work on RPostgres.

The test failure seems to be a withr problem that should be corrected if you trigger a new test run, e.g. with an empty commit to the branch.

zozlak:
Great, thanks!

krlmlr:
Thanks!

PR #313, opened 2021-07-19
change docs URLs to "current" version and https

dpprdan:
This updates the URLs to the PostgreSQL docs to the "current" version and https scheme. 
I wasn't sure about the URLs in the cpp files. Do you want them to point at the version at the time of writing?

### DIFF
--- a/R/PqConnection.R
+++ b/R/PqConnection.R
@@ -136,21 +136,21 @@ setMethod("dbGetInfo", "PqConnection", function(dbObj, ...) {
 #'   Note that this argument can only contain the database name, it will not
 #'   be parsed as a connection string (internally, `expand_dbname` is set to
 #'   `false` in the call to
-#'   [`PQconnectdbParams()`](https://www.postgresql.org/docs/9.6/static/libpq-connect.html)).
+#'   [`PQconnectdbParams()`](https://www.postgresql.org/docs/current/libpq-connect.html)).
 #' @param user,password User name and password. If `NULL`, will be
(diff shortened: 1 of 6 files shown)


krlmlr:
Thanks. Will CRAN's URL check complain?

dpprdan:
Cannot speak for CRAN 😉, but not according to {urlchecker}:

```r
> library(urlchecker)
> url_check()
√ All URLs are correct!
```

`curl` returns a HTTP 200 status

```bash
$ curl -I  'https://www.postgresql.org/docs/current/libpq-connect.html'
HTTP/2 200
server: nginx
date: Tue, 07 Sep 2021 12:48:20 GMT
content-type: text/html; charset=utf-8
x-frame-options: DENY
content-security-policy: default-src 'self' ; img-src * data: ; script-src 'unsafe-eval' 'self' www.google-analytics.com ssl.google-analytics.com www.googletagmanager.com tagmanager.google.com data: ; connect-src 'self' www.google-analytics.com ssl.google-analytics.com ; media-src 'self' ; style-src 'self' fonts.googleapis.com tagmanager.google.com 'unsafe-inline' ; font-src 'self' fonts.gstatic.com data: ; frame-ancestors 'none' ; report-uri https://csp-report.postgresql.org/
x-xss-protection: 1; mode=block
vary: Accept-Encoding
x-varnish: 24429296 19814862
age: 14738
via: 1.1 varnish (Varnish/6.0)
strict-transport-security: max-age=31536000
accept-ranges: bytes
```

krlmlr:
Thanks!

PR #337, opened 2021-09-19
Fix interrupt checking

### LINKED ISSUE #336: dbConnect(check_interrupt = TRUE) broken?
The interrupt seems to be caught by `select()` on the socket, not by `Rcpp::checkUserInterrupt()` as intended by the code. This leads to an error message `Rcpp::exception: select() failed with error code 4` (where 4 means `EINTR`), and the subsequent cleanup (in an overarching try-catch?) still waits for the query to complete.

Same on macOS and Linux. Not tested on Windows.

``` r
session <- callr::r_session$new()

session$run(function() {
  library(RPostgres)
  .GlobalEnv$conn <- dbConnect(Postgres(), check_interrupts = TRUE)
  invisible()
})
#> NULL
session$call(function() {
  tryCatch(
    print(dbGetQuery(.GlobalEnv$conn, "SELECT pg_sleep(2)")),
    error = identity
  )
})

session$poll_process(100)
#> [1] "timeout"
session$read()
#> NULL
session$interrupt()
#> [1] TRUE
# Should take much less than 1.9 seconds
system.time(print(session$poll_process(2000)))
#> [1] "ready"
#>    user  system elapsed 
#>   0.000   0.001   1.904
# Should return a proper error message
session$read()
#> $code
#> [1] 200
#> 
#> $message
#> [1] "done callr-rs-result-f8d028852a34"
#> 
#> $result
#> <Rcpp::exception: select() failed with error code 4>
#> 
#> $stdout
#> [1] ""
#> 
#> $stderr
#> [1] ""
#> 
#> $error
#> NULL
#> 
#> attr(,"class")
#> [1] "callr_session_result"
```

<sup>Created on 2021-09-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

krlmlr:
Closes #336.

``` r
session <- callr::r_session$new()

session$run(function() {
  library(RPostgres)
  .GlobalEnv$conn <- dbConnect(Postgres(), check_interrupts = TRUE)
  invisible()
})
#> NULL
session$call(function() {
  tryCatch(
    print(dbGetQuery(.GlobalEnv$conn, "SELECT pg_sleep(2)")),
    error = identity
  )
})

session$poll_process(100)
#> [1] "timeout"
session$read()
#> NULL
session$interrupt()
#> [1] TRUE
# Should take much less than 1.9 seconds
time <- system.time(session$poll_process(2000))
print(time)
#>    user  system elapsed 
#>   0.000   0.000   0.001
# Should return a proper error message
out <- session$read()
out
#> $code
#> [1] 200
#> 
#> $message
#> [1] "done callr-rs-result-106b8135f5fa5"
#> 
#> $result
#> <Rcpp::exception: Failed to fetch row: ERROR:  canceling statement due to user request
#> >
#> 
#> $stdout
#> [1] ""
#> 
#> $stderr
#> [1] ""
#> 
#> $error
#> NULL
#> 
#> attr(,"class")
#> [1] "callr_session_result"
```

<sup>Created on 2021-09-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -33,9 +33,12 @@ jobs:
       fail-fast: false
       # Ensure that the "cancel" workflow gets a chance to run quickly, even if we just pushed
       # Need to figure out how to smoke-test
-      max-parallel: 8
+      max-parallel: 6
       matrix:
         config:
+          - { os: macOS-latest,                 r: 'release' }
+          - { os: windows-latest,               r: 'release', rspm: "https://packagemanager.rstudio.com/all/latest" }
+          - { os: windows-latest,               r: 'oldrel',  rspm: "https://packagemanager.rstudio.com/all/latest" }
           - { os: ubuntu-, os-version: 20.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest" }
           - { os: ubuntu-, os-version: 18.04,   r: 'release', rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest" }
           - { os: ubuntu-, os-version: 18.04,   r: 'devel',   rspm: "https://packagemanager.rstudio.com/cran/__linux__/bionic/latest", http-user-agent: "R/4.0.0 (ubuntu-18.04) R (4.0.0 x86_64-pc-linux-gnu x86_64 linux-gnu) on GitHub Actions" }
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -33,7 +33,7 @@ jobs:
       fail-fast: false
       # Ensure that the "cancel" workflow gets a chance to run quickly, even if we just pushed
       # Need to figure out how to smoke-test
-      max-parallel: 7
+      max-parallel: 8
       matrix:
         config:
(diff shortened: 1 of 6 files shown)


krlmlr:
CC @zozlak: looks like we need to check for `EINTR` and also actively cancel the query. Done here, let me know if you see any issues with this approach.

zozlak:
> and also actively cancel the query

Sure, sorry for missing it.

> we need to check for EINTR

While we definitely should, it looks like what we should do with it depends on the platform (sic!).

* [The Windows documentation](https://docs.microsoft.com/en-us/windows/win32/winsock/windows-sockets-error-codes-2) suggests the `WSAEINTR` should be treated as a connection cancellation.
  * By the way it looks like it can only happen if Windows Sockets v1.1 are in use (see the deprecation note on [this page](https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-wsacancelblockingcall)) but as handling is straightforward I would say there is no harm in handling it even if it can't really pop up on modern Windows.
* [Linux man on select](https://man7.org/linux/man-pages/man2/select.2.html#NOTES) states that the `EINTR` should be interpreted as "your select() call has been interrupted, try again". Which is in contrary to your current implementation.

PR #356, opened 2021-11-07
Use Postgres 14

krlmlr:
to fix CI.

Older versions not supported on Windows.

### DIFF
--- a/.github/workflows/R-CMD-check.yaml
+++ b/.github/workflows/R-CMD-check.yaml
@@ -140,7 +140,7 @@ jobs:
       # https://github.com/ankane/setup-mariadb/issues/2
       - uses: ankane/setup-postgres@v1
         with:
-          postgres-version: 13
+          postgres-version: 14
 
       - name: Create database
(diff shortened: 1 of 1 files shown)


dpprdan:
Ha, I've been looking into this [as well](https://github.com/dpprdan/RPostgres/pull/1). 

ICYMI: You could also omit the [`with: postgres-version`](https://github.com/r-dbi/RPostgres/blob/9a219ea256bd1a48ac51d3bc6f6049a186c42603/.github/workflows/R-CMD-check.yaml#L124-L125), which will then use the [latest version](https://github.com/ankane/setup-postgres/pull/5) supported by the setup-postgres action (which might differ per OS). For macOS this was still Postgres v13 [as of two days ago](https://github.com/ankane/setup-postgres/pull/4). 
That way the workflow doesn't have to be kept in sync manually with what happens upstream; assuming you want to test on the latest available version anyway.

krlmlr:
Superseded.

Issue #312, opened 2021-07-06
Re-licensing RPostgres


hadley:
We are systematically re-licensing tidyverse, r-lib, and now (where possible) r-dbi packages to use the MIT license, to make our package licenses as clear and permissive as possible. To do so, we need the approval of all copyright holders, which I have found by reviewing contributions from all all non-RStudio [contributors](https://github.com/r-dbi/RPostgres/graphs/contributors). @Altons, @ateucher, @baderstine, @baileych, @etiennebr, @harvey131, @imlijunda, @JackStat, @jaredk-porch, @jeroen, @lentinj, @mmuurr, @oz-r, @robertzk, @thrasibule, @trafficonese, @troels, @TSchiefer, @zozlak, would you permit us to re-license RPostgres with the MIT license? If so, please comment "I agree" below.
​
* [x] [@Altons](https://github.com/r-dbi/RPostgres/commits?author=Altons)  
* [x] [@ateucher](https://github.com/r-dbi/RPostgres/commits?author=ateucher)  
* [x] [@baderstine](https://github.com/r-dbi/RPostgres/commits?author=baderstine)  
* [x] [@baileych](https://github.com/r-dbi/RPostgres/commits?author=baileych)  
* [x] [@etiennebr](https://github.com/r-dbi/RPostgres/commits?author=etiennebr)  
* [x] [@harvey131](https://github.com/r-dbi/RPostgres/commits?author=harvey131)  
* [x] [@imlijunda](https://github.com/r-dbi/RPostgres/commits?author=imlijunda)  
* [ ] [@JackStat](https://github.com/r-dbi/RPostgres/commits?author=JackStat)  
* [ ] [@jaredk-porch](https://github.com/r-dbi/RPostgres/commits?author=jaredk-porch)  
* [x] [@jeroen](https://github.com/r-dbi/RPostgres/commits?author=jeroen)  
* [x] [@lentinj](https://github.com/r-dbi/RPostgres/commits?author=lentinj)  
* [x] [@mmuurr](https://github.com/r-dbi/RPostgres/commits?author=mmuurr)  
* [x] [@oz-r](https://github.com/r-dbi/RPostgres/commits?author=oz-r)  
* [x] [@robertzk](https://github.com/r-dbi/RPostgres/commits?author=robertzk)  
* [x] [@thrasibule](https://github.com/r-dbi/RPostgres/commits?author=thrasibule)  
[...]

PR #358, opened 2021-11-23
Allow to find table in temporary schema for Redshift

galachad:
Issue:
```
> DBI::dbWriteTable(con, "iris", iris, temporary = TRUE, overwrite = TRUE)
> DBI::dbWriteTable(con, "iris", iris, temporary = TRUE, overwrite = TRUE)
 Error: Failed to fetch row: ERROR:  Relation "iris" already exists 
```

Using 
```
(SELECT 1 AS nr, current_schema() AS table_schema) ttt
```
is nice solution but it do not return temporary schema same as `current_schemas(true)`.

This change works correctly with temp schema after creating temporary table in Redshift.


krlmlr:
Thanks, this is amazing! A slew of tests now pass on Redshift, 4025336. The only thing that seems to be missing now is `dbRemoveTable(temporary = TRUE)` . Any idea how to achieve this?

krlmlr:
Also, I'd appreciate a brief review of my CTE rewrite.

galachad:
The changes look great. I also added support for `dbRemoveTable(temporary = TRUE)` in Redshift, please check.

krlmlr:
Thanks a lot, this looks great! A few remarks:

- Do the schema names always start with `pg_temp_` ? The pattern should be `pg_temp_%` in this case perhaps.
- If we can't fetch the schema name because it doesn't exist, we should raise an error if `fail_if_missing` is `TRUE` .
- Should we cache the schema name so that it is fetched only once?
- Can we extract the logic of obtaining the schema name into a function? This function can return `NULL` if the temp schema doesn't exist yet, or perhaps take a `fail_if_missing` argument.

Would you like to take another stab?

galachad:
Sure

- Yes it's `pg_temp_%`, I think Redshift is managing temp schema just by adding access to one of existing schemas and clean up after finish connection. Redshift doesn't has alias like in Postgres.
```
SELECT * FROM pg_catalog.pg_namespace WHERE nspname like 'pg_temp_%'
```
- I added raising error when `fail_if_missing` is `TRUE` (for Redshift).
- That's nice idea. Added.
- Extracted to `find_temp_schema` function.

krlmlr:
```
══ Results ══════════════════════════════════════════════════════════════════════════════════════
Duration: 385.5 s

── Skipped tests  ───────────────────────────────────────────────────────────────────────────────
• tweak: omit_blob_tests (7)

[ FAIL 0 | WARN 0 | SKIP 7 | PASS 6386 ]
```

With all tests enabled. Thanks a lot!

krlmlr:
One more thing: do you agree to contribute under the MIT license? Please indicate your consent or disapproval in #312. Thanks!

PR #359, opened 2021-11-23
Handling table path as character

galachad:
**Case**:
```
dbWriteTable(con, "private_schema.iris", iris)
```

Less experienced users may use the above code to write the table to `"private"` schema.
This will in fact write table named `private.schema.iris` to `"public"` schema, which may result with accidental leak of data. It also simplify usability for this group of users. 

**Issue**:
`dbQuoteIdentifier`  is not handling full table identifier as character .

**Example**:
Execution of `dbWriteTable(con, "private_schema.iris", iris)` will save the table to `"current_schema"."private_schema.iris"`.

This PR is a proposition to add an option to unquote the identifier while calling `dbQuoteIdentifier`.

It simply allows read/write methods from/to `"private_schema"."iris"`.

**Description**:
- `unquote_id` the parameter that can enable the feature (specified while setting up the connection) - setting to `TRUE` converts character to the identifier before being quoted.

Because `dbQuoteIdentifier` doesn't store the information about type of the data (schema, table, fields) I've had to detect it with the usage of `match.call`: `!grepl("names\\((.*?)\\)", as.character(match.call())[[3]])`.
This allows to handle cases where table fields contain `.` character in names.

### DIFF
--- a/R/PqConnection.R
+++ b/R/PqConnection.R
@@ -12,7 +12,8 @@ setClass("PqConnection",
     bigint = "character",
     timezone = "character",
     timezone_out = "character",
-    typnames = "data.frame"
+    typnames = "data.frame",
+    unquote_id = "logical"
   )
 )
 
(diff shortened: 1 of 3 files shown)


krlmlr:
Thanks, good point about inexperienced users. Defining unquoting behavior on the connection level may work but doesn't feel quite right here.

Ideally, we would require users to quote all identifiers, maybe in dbi3. For now, how about requiring users to quote identifiers if they contain a dot?

krlmlr:
Thanks again for your contribution. I have opened an issue in the dbi3 repo so that this can be tackled in a future iteration.

PR #366, opened 2021-12-03
Use Postgres 14 (older versions not supported on Windows)

krlmlr:
with new workflow.


krlmlr:
Seeing the same failures locally on the M1 mac 🎆

PR #375, opened 2021-12-06
Bulk insertion for sqlAppendTable.

galachad:
This is another PR important for Redshift compatibility.

By default it's not possible to send large data frame using SQL method as statement is too large.
```
Error: Failed to prepare query: ERROR:  Statement is too large. Statement Size: 26100068 bytes. Maximum Allowed: 16777216 bytes
```
The stdin copy method is not supported on Redshift so appending data  using sql statement(`sqlAppendTable`). 

This simple implementation just split data frame into smaller pieces. By editing `max_bulk_lenght` user has possibility to adjust number of row in bulk to achieve better performance or to avoid `Statement is too large.` error.

I happy to discuss how we can handle splitting data frame to achieve best performance and do not raise statement length error. cc: @krlmlr

Examples:

```
# Standard example
n <- 200000
testdat <- data.frame(
  id = 1:n,
  num1 = rnorm(n),
  num2 = rnorm(n),
  num3 = rnorm(n),
  num4 = rexp(n),
  char1 = stringi::stri_rand_strings(n, 20),
  date  = lubridate::ymd("1990-01-01") + lubridate::days(round(runif(n, 0, 10000))),
  datetime = lubridate::ymd_hms("1990-01-01T00:00:00") + lubridate::seconds(round(runif(n, 0, 1000000000))),
  fct = factor(sample(c("A", "B", "C", "D"), size = n, replace = TRUE))
)
dbWriteTable(con, "test_data", testdat, overwrite = TRUE, temporary = TRUE)
```

```
# Every cell contains long character
n <- 100000
testdat <- data.frame(
  char1 = stringi::stri_rand_strings(n, 256),
  char2  = stringi::stri_rand_strings(n, 256),
  char3 = stringi::stri_rand_strings(n, 256),
  char4 = stringi::stri_rand_strings(n, 256)
)
dbWriteTable(con, "test_data", testdat, overwrite = TRUE, temporary = TRUE, max_bulk_lenght = 50000)
```

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -37,6 +37,9 @@
 #'   a single SQL string. This is slower, but always supported.
 #'   The default maps to `TRUE` on connections established via [Postgres()]
 #'   and to `FALSE` on connections established via [Redshift()].
+#' @param bulk_insertion If `TRUE` and copy is `FALSE` split value into
+#' smaller pieces and load them one by one.
+#' @param max_bulk_lenght Defines maximum number of cells to send in one bulk.
 #'
 #' @examplesIf postgresHasDefault()
 #' library(DBI)
@@ -60,7 +63,8 @@ NULL
 #' @rdname postgres-tables
 setMethod("dbWriteTable", c("PqConnection", "character", "data.frame"),
   function(conn, name, value, ..., row.names = FALSE, overwrite = FALSE, append = FALSE,
-           field.types = NULL, temporary = FALSE, copy = NULL) {
+           field.types = NULL, temporary = FALSE, copy = NULL, bulk_insertion = TRUE,
+           max_bulk_lenght = 600000) {
 
     if (is.null(row.names)) row.names <- FALSE
     if ((!is.logical(row.names) && !is.character(row.names)) || length(row.names) != 1L)  {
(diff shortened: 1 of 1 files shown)


krlmlr:
Thanks. The AWS documentation suggests that `COPY` from an S3 bucket is the fastest way to load data. It seems easy enough for clients to split the input data into chunks if they really want to use `dbAppendTable()` . I'm slightly in favor of keeping things as they are and point to alternative solutions instead.

https://github.com/sicarul/redshiftTools seems to provide tooling (not verified) but is now archived on CRAN because aws.s3 is archived.

galachad:
In the internal package we already have the implementation method that allows upload dataset using S3 but it not available in all infrastructure or for all users. Having this functionality even optional will be super useful and simplify working with Redshift. 

This is similar behavior that is used in `pandas.DataFrame.to_sql`  where it's possible to define `chunksize = 20000` and method = 'multi'`. The change do not impact Postgres connection as Postgres is using `COPY STDIN` method.

We just wanted to avoid explaining users that they need to split dataset into smaller chunks.

Do you see any chance that this change will be added into RPostgres?

krlmlr:
Thanks, good point. I don't see this change in RPostgres for two reasons:

- `dbAppendTable()` should do one thing, and do it well.
- It feels that this functionality might be useful for other backends too. 
 
We'll need some time to flesh it out, I can't accept this PR at this time. I have posted a variant to StackOverflow that avoids copying the data: https://stackoverflow.com/a/70405875/946850.

I greatly appreciate your contributions.

PR #385, opened 2021-12-21
Issue 344 tests

zozlak:
* Fixed timing in the issue #244 test (the issue affected queries running **longer** than 1 second)
* Query interrupt test extended with a connection PID-based check if the query has been cancelled also on the server.

### DIFF
--- a/tests/testthat/test-checkInterrupts.R
+++ b/tests/testthat/test-checkInterrupts.R
@@ -20,6 +20,8 @@ test_that("check_interrupts = TRUE interrupts immediately (#336)", {
   session$run(function() {
     library(RPostgres)
     .GlobalEnv$conn <- postgresDefault(check_interrupts = TRUE)
+    .GlobalEnv$connPid <- dbGetQuery(.GlobalEnv$conn, "SELECT pg_backend_pid() AS pid")$pid
+    .GlobalEnv$checkConn <- postgresDefault()
     invisible()
   })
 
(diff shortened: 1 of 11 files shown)


tests/testthat/test-checkInterrupts.R
```diff
@@ -1,9 +1,9 @@
 context("checkInterrupts")
 
-test_that("check_interrupts = TRUE works with queries < 1 second (#244)", {
+test_that("check_interrupts = TRUE works with queries > 1 second (#244)", {
```
krlmlr, line 3: The original test is good. Can you please create a new test?

Should the test condition be `expect_lt(time[["elapsed"]], 0.9)` then?

krlmlr (commented):
Thanks. The tests are failing now.

#244 was about short queries that still waited for one second. What is the new test testing?

tests/testthat/test-checkInterrupts.R
zozlak: If we claim this is a test for #244, it should definitely test a query longer than 1 second because the #244 was caused by a fact that the structure taken by the `select()` requires different reinitialization under linux and windows, after the `select()` timeout is reached (and we hardcode the timeout to 1 second).

Of course I can also add another test with query shorter than 1 second, just then we shouldn't claim it tests #244.

krlmlr: Maybe the reference to #244 is wrong, I'd need to take a closer look. Either way, can you please add a second test?

krlmlr: The new test is failing, btw.

zozlak: Sure, restored the old removing the reference to #244 and added the > 1 second one with the reference to #244 and fixed check condition.

zozlak: Hm, the test CI failed on the "update snapshots" step - I have no idea what does it mean.
When I run tests locally all pass but `(test-checkInterrupts.R:50:3): check_interrupts = TRUE interrupts immediately (#336)` (which means the query isn't cancelled on the server side when the interrupt breaks `dbGetQuery()`).

krlmlr:
Thanks!

PR #389, opened 2022-01-05
adding observer for RStudio Connection

meztez:
In case anyone need something similar.

I like to be able to use the connections pane in RStudio.

Comments welcome.


krlmlr:
Thanks, appreciated. Many packages now implement the Connections Contract: https://github.com/search?q=org%3Acran+connectionObserver&type=code . There's also an rscontract package, and it's documented here: <https://rstudio.github.io/rstudio-extensions/connections-contract.html>.

I wonder if we could roll a solution in the DBI package so that all backends can benefit. The icon seems the most challenging, but could be made available either as a slot in the connection object, or as a new method?

PR #390, opened 2022-01-29
feat!: `dbListObjects()` only allows `Id()` objects as `prefix` argument

dpprdan:
This is an attempt to only allow `Id()` objects as `prefix` argument to `dbListObjects()` as mentioned in https://github.com/r-dbi/RPostgres/pull/372#issuecomment-1003884250.

> It looks to me like `dbListObjects()` should always be called with an `Id()`, never with a `SQL()`

prefix can now be a list with multiple `Id` objects, so querying all tables in all schemata a user has access to like this works:

```r
con <- postgresDefault()
tbl <- dbListObjects(con)$table
head(dbListObjects(con, tbl))
dbDisconnect(con)
```

I am not sure, though, whether this was the intent for using `vlapply()` and `vcapply()` here previously or whether I am missing good reasons against making `prefix` in `dbListObjects()` vectorised?

Wouldn't it be cleaner though, to only have an `Id` method (and one for `MISSING`) for `dbListObjects()` instead of an `ANY` one? But that would require those changes be made to DBI (and other DBI-related packages), as well. I am probably missing something there as well.


krlmlr:
Re-triggered failing job.

krlmlr:
Thanks!

Issue #394, opened 2022-04-29
connection fails with error: view 'pg_type' not found


pedrobtz:
When connecting to a Denodo (Postgress) database one gets error `view 'pg_type' not found`

The error happens at line
https://github.com/r-dbi/RPostgres/blob/e9f0366742217b9337979e914215419000970629/R/dbConnect_PqDriver.R#L111

If ones debugs function call `dbConnect_PqDriver`, the connection is established and one can query correctly other tables.

The  python packages `pydobc` and `psycopg2` work okay.

krlmlr:
Thanks. This package assumes that a view or table named `pg_type` exists, which doesn't seem to be the case here. Would you like to submit a PR that fills `conn@typnames` with an empty two-column data frame if `dbGetQuery()` is unsuccessful?

pedrobtz:
many thanks.

PR #395, opened 2022-04-29
conn@typnames with an empty two-column data frame

pedrobtz:
Issue #394 
fills conn@typnames with an empty two-column data frame if dbGetQuery() is unsuccessful


krlmlr:
Thanks!

pedrobtz:
thanks, danke.

Issue #388, opened 2021-12-30
`dbListObjects()` does not list empty schemata


dpprdan:
`dbListObjects()` does not list empty schemata, i.e. a schema without a table.

``` r
library(RPostgres)
con <- postgresDefault()
dbExecute(con, paste0('CREATE SCHEMA "Robert"'))
#> [1] 0
dbListObjects(con)
#>                              table is_prefix
#> 1 <Id> schema = information_schema      TRUE
#> 2         <Id> schema = pg_catalog      TRUE
```

The schema is listed in `information_schema.schemata` or `pg_catalog.pg_namespace` (the (empty) `public` schema is also not listed by `dbListObjects()` BTW)

``` r
dbGetQuery(con, 'SELECT schema_name FROM information_schema.schemata;')
#>          schema_name
#> 1           pg_toast
#> 2         pg_catalog
#> 3             public
#> 4 information_schema
#> 5             Robert
dbGetQuery(con, 'SELECT * FROM pg_catalog.pg_namespace;')
#>     oid            nspname nspowner                              nspacl
#> 1    99           pg_toast       10                                <NA>
#> 2    11         pg_catalog       10  {postgres=UC/postgres,=U/postgres}
#> 3  2200             public       10 {postgres=UC/postgres,=UC/postgres}
#> 4 13111 information_schema       10  {postgres=UC/postgres,=U/postgres}
#> 5 16417             Robert       10                                <NA>
```

The schema is also listed by `dbListObjects()` after adding a table

``` r
table_qid <- dbQuoteIdentifier(con, Id(schema = 'Robert', table = "mtcars"))
dbWriteTable(con, table_qid, mtcars)
dbListObjects(con)
#>                              table is_prefix
#> 1 <Id> schema = information_schema      TRUE
#> 2             <Id> schema = Robert      TRUE
#> 3         <Id> schema = pg_catalog      TRUE

# clean-up
dbExecute(con, paste0('DROP TABLE ', table_qid))
#> [1] 0
dbExecute(con, paste0('DROP SCHEMA "Robert"'))
#> [1] 0
dbDisconnect(con)
```

[...]

PR #413, opened 2022-11-29
chore: refactor `dbListTables()` et al.

dpprdan:
I would like to take another stab at #251 and this some preparatory work for that.

# Background

Materialized views are currently not returned by `dbListTables()` et al (#251). Materialized Views are not included in `INFORMATION_SCHEMA.tables` (see [this thread](https://www.postgresql.org/message-id/flat/CAH7T-ao6ece1mgCsCvsE04W59ZZJP9gGXVK97wkUzRV5gsDqQQ%40mail.gmail.com) for why they are not. Tl;dr: "They are not defined by the SQL standard.")

> The information schema consists of a set of views that contain information about the objects defined in the current database. 
> The information schema is defined in the SQL standard and can therefore be expected to be portable and remain stable — unlike the system catalogs, which are specific to PostgreSQL and are modeled after implementation concerns. 
> The information schema views do not, however, contain information about PostgreSQL-specific features; to __inquire about those you need to query the system catalogs__ or other PostgreSQL-specific views.
> https://www.postgresql.org/docs/current/information-schema.html (emphasis mine).

Sidenote: Most if not all tables in information_schema are just views of the system catalogs, see e.g. 

```
SELECT definition FROM pg_views 
WHERE schemaname = 'information_schema' AND viewname = 'tables';
```

Since mviews are PostgreSQL-specific features we need to query the system catalogs. In particular, we need to use [`pg_class`](https://www.postgresql.org/docs/current/catalog-pg-class.html)/[`pg_namespace`](https://www.postgresql.org/docs/current/catalog-pg-namespace.html) instead of `INFORMATION_SCHEMA.tables`. 

However, in order to retain support for Redshift, we need to keep the `INFORMATION_SCHEMA` queries as well.

# What happens in this PR?

The above affects the queries that underly the `dbListTables()`, `dbExistsTable()`, `dbListObjects()`, and `dbListFields()` functions.  

`dbListTables()`, `dbExistsTable()` and `dbListObjects()` do essentially the same thing, namely call `INFORMATION_SCHEMA.tables`. 
In order to make the abovementioned adjustments easier and keep duplicated code at a minimum, I refactored those functions to use a common core function `list_tables()`, which returns the SQL code to query `INFORMATION_SCHEMA.tables`. The (alternative) queries to `pg_class` (to be added in another PR) will then live here.

I refactored the `find_table()` and `list_fields()` functions that underly `dbListFields()` in a similar fashion. 

<details>
<summary>Details</summary>

`dbListFields()`: find the first table on the search path

```sql
SELECT column_name FROM (
  SELECT *, rank() OVER (ORDER BY nr) AS rnr FROM (
    SELECT nr, schemas[nr] AS table_schema FROM (
      SELECT *, generate_subscripts(schemas, 1) AS nr FROM (
        SELECT current_schemas(true) AS schemas) t
      ) tt WHERE schemas[nr] <> 'pg_catalog'
    ) ttt 
  INNER JOIN INFORMATION_SCHEMA.columns USING (table_schema) WHERE table_name = [table_name]
) tttt 
WHERE rnr = 1 ORDER BY ordinal_position
```

can be re-written as (with 2 instead of 4 nested queries)

```sql
SELECT column_name FROM (
  SELECT *, rank() OVER (ORDER BY schema_nr) AS schema_rank FROM (
    SELECT * FROM unnest(current_schemas(true)) WITH ORDINALITY AS "tbl"("table_schema","schema_nr") 
    WHERE "table_schema" != 'pg_catalog') t 
  INNER JOIN INFORMATION_SCHEMA.columns USING (table_schema) 
  WHERE table_name = [table_name]) tt 
WHERE schema_rank = 1 
ORDER BY ordinal_position
```

</details>

Essentially I applied similar query code here as in #261, but calling `information_schema` instead of the system catalogs.
I will open another PR to incorporate the queries to the system catalogs.

It is probably easiest to review this by going through the individual commits.

I do not have access to Redshift, so I was not able to test this PR there. There might be small adjustments necessary (see below). Is it somehow possible for me to test on Redshift (e.g. via GHA)? Or how does this work in general?

# Related issues

https://github.com/r-dbi/RPostgres/issues/251
https://github.com/r-dbi/RPostgres/pull/261

This does not address #388. I think it is best to address this after #251 is resolved.

#390 is also related, but relatively trivial to apply before or after this.

### DIFF
--- a/R/tables.R
+++ b/R/tables.R
@@ -120,26 +120,72 @@ db_append_table <- function(conn, name, value, copy, warn) {
   nrow(value)
 }
 
-exists_table <- function(conn, id) {
-  name <- id@name
+list_tables <- function(conn, where_schema = NULL, where_table = NULL, order_by = NULL) {
 
   query <- paste0(
-    "SELECT COUNT(*) FROM ",
-    find_table(conn, name)
+    # information_schema.table docs: https://www.postgresql.org/docs/current/infoschema-tables.html
+    "SELECT table_schema, table_name \n",
+    "FROM information_schema.tables \n",
+    "WHERE TRUE \n" # dummy clause to be able to add additional ones with `AND`
   )
 
-  dbGetQuery(conn, query)[[1]] >= 1
+  if (is.null(where_schema)) {
+    # `true` in `current_schemas(true)` is necessary to get temporary tables
+    query <- paste0(
+      query,
+      "  AND (table_schema = ANY(current_schemas(true))) \n",
+      "  AND (table_schema <> 'pg_catalog') \n"
+    )
+  } else {
+    query <- paste0(query, "  AND ", where_schema)
+  }
+
+  if (!is.null(where_table)) query <- paste0(query, "  AND ", where_table)
+
+  if (!is.null(order_by)) query <- paste0(query, "ORDER BY ", order_by)
+
+  query
 }
 
-find_table <- function(conn, id, inf_table = "tables", only_first = FALSE) {
+exists_table <- function(conn, id) {
+  name <- id@name
+  stopifnot("table" %in% names(name))
+  table_name <- dbQuoteString(conn, name[["table"]])
+  where_table <- paste0("table_name = ", table_name, " \n")
+
+  if ("schema" %in% names(name)) {
+    schema_name <- dbQuoteString(conn, name[["schema"]])
+    where_schema <- paste0("table_schema = ", schema_name, " \n")
+  } else {
+    where_schema <- NULL
+  }
+  query <- paste0(
+    "SELECT EXISTS ( \n",
+    list_tables(conn, where_schema = where_schema, where_table = where_table),
+    ")"
+  )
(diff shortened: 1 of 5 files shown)


R/tables.R
```diff
@@ -156,25 +206,31 @@ find_table <- function(conn, id, inf_table = "tables", only_first = FALSE) {
     )
     only_first <- FALSE
```
dpprdan, line 207: This override was already present in `find_table()` (but `find_table()` was generally called with `only_first = TRUE` by `list_fields()`).

I think this is a remainder of the first two commits in #326, where indeed `only_first` isn't necessary, because that used current_schema() (not "schema_s"). 

It is necessary now though because there can be multiple schemas on the search path on Redshift as well (I think).

krlmlr:
Conflicts here, too.

krlmlr:
Thanks!

PR #414, opened 2022-12-15
list materialized views in Postgres

### LINKED ISSUE #251: materialized views in dbListTables()

dpprdan:
Draft

uses system catalogs instead of information schema under the hood for `dbList*()` and `dbExistsTable()` to be able to list information about materialized views. Adds tests for materialized views.

extends #413, improved version of #261, closes #251


krlmlr:
This now has conflicts.

R/dbListObjects_PqConnection_ANY.R
```diff
@@ -14,11 +14,11 @@ dbListObjects_PqConnection_ANY <- function(conn, prefix = NULL, ...) {
     }
     query <- paste0(
       "SELECT ", null_varchar, " AS schema, table_name AS table FROM ( \n",
-      list_tables(conn = conn, order_by = "table_type, table_name"),
+        list_tables(conn = conn, order_by = "table_type, table_name"),
```
krlmlr, line 17: styler won't like it. Maybe use braces or parens to enforce indent here?

R/tables.R
```diff
@@ -252,6 +301,29 @@ list_fields <- function(conn, id) {
   fields
 }
 
+list_fields_from_system_catalog <- function(conn, id) {
+  if (exists_table(conn, id)) {
+    # we know from exists_table() that id@name["table"] exists
+    # and the user has access priviledges
+    tname_str <- stats::na.omit(id@name[c("schema", "table")])
```
krlmlr, line 308: `Id()` now has unnamed components only. How to deal with this here?

R/dbConnect_PqDriver.R
```diff
@@ -60,7 +67,8 @@
 dbConnect_PqDriver <- function(drv, dbname = NULL,
                                host = NULL, port = NULL, password = NULL, user = NULL, service = NULL, ...,
                                bigint = c("integer64", "integer", "numeric", "character"),
-                               check_interrupts = FALSE, timezone = "UTC", timezone_out = NULL) {
+                               check_interrupts = FALSE, system_catalogs = TRUE,
```
krlmlr, line 70: The new argument name is not very clear, perhaps use something like `meta = c("system", "information_schema")` ?

I'm not in love with `meta` either, can you think of a better name?

krlmlr (commented):
Thanks. I now disentangled and merged most preparatory PRs. Should we take another stab here?

dpprdan:
> Thanks. I now disentangled and merged most preparatory PRs. Should we take another stab here?

🥳 I'll look into it, but again, it might take a couple of days until I'll get around to it.

R/tables.R
dpprdan: We have more places where we rely on named `Id()`s, I am afraid, also in current main, which are not caught by current tests. 😬 
I can take a stab at https://github.com/r-dbi/DBItest/issues/367 if you want? Either way, I'll try to amend the tests for matviews first and work my way up from there:

https://github.com/dpprdan/RPostgres/blob/a3dfc02dff54b84e944efe2b68c9e4e8cc447fd8/tests/testthat/test-list_matviews.R#L35-L40

Which `name` variants are (supposed to be) supported now? 
- those [formerly allowed already](https://github.com/r-dbi/DBI/blob/6382f918a56b78eb832348e0eedb538a6721286a/R/dbReadTable.R#L22-L27)
- `Id("schema_name", "table_name")`
- `I("schema_name.table_name")` [as well](https://www.tidyverse.org/blog/2024/04/dbplyr-2-5-0/)?

Relatedly: Is https://github.com/r-dbi/DBItest/issues/340 relevant for RPostgres?

R/dbListObjects_PqConnection_ANY.R
dpprdan: I'm fine with not having an extra indent here and prefer that to extra braces or parentheses.

If I run `styler::style_package()` with `tidyverse_style()` it'll style a lot more files than those relevant for this PR. So I'll just run styler on the latter and leave the others for a separate PR?!

R/dbConnect_PqDriver.R
dpprdan: `meta` seems quite broad and therefore not very clear as well, so how about `meta_schema`?

If we'd want to stick with a boolean, it could also be `information_schema`, but it's probably better to use a character argument to keep it extendable? 

(Just documenting some related info:)
PostgreSQL calls the tables with the system information [`System Catalogs`](https://www.postgresql.org/docs/current/catalogs.html). There are also [`System Views`](https://www.postgresql.org/docs/current/views.html), however, like `pg_matviews`. Both the Catalogs and the Views reside in the `pg_catalog` schema.

Redshift also has [Materialized Views](https://docs.aws.amazon.com/redshift/latest/dg/materialized-view-overview.html) (out of scope for this PR). Their meta info resides (mostly) in [`ST_MV_INFO`](https://docs.aws.amazon.com/redshift/latest/dg/r_STV_MV_INFO.html), one of the ["system tables and views"](https://docs.aws.amazon.com/redshift/latest/dg/cm_chap_system-tables.html). The ["System catalog tables"](https://docs.aws.amazon.com/redshift/latest/dg/c_intro_catalog_views.html), which seem to be legacy tables from the PostgreSQL v8 times, are a part of those as well. 

Long story short: `meta_schema = c("system", "information_schema")` seems to be the most fitting argument to specify which schema should be queried for information on tables, (materialized) views and columns.

PR #416, opened 2023-01-18
Fix warnings sprintf

Antonov548:
- remove `sprintf` from boost version without buffer size argument
- fix boost warning related to `assert` - https://github.com/boostorg/mpl/pull/34


krlmlr:
Thanks, fixed by #417.

PR #428, opened 2023-03-18
chore: Avoid bashisms warning on Ubuntu 22.04

pachadotdev:
hi @krlmlr 
this comes from some changes I proposed here https://github.com/pachadotdev/RPostgres/blob/cpp11/configure


krlmlr:
Thanks!

PR #433, opened 2023-04-07
chore: tweak driver docs

dpprdan:
This tweaks the `drv` docs a little to make it clearer that there are two alternatives and when to use which.

Motivation: "drv: Should be set to [RPostgres::Postgres()] to use the \pkg{RPostgres} package." does not sound correct anymore.


krlmlr:
Thanks!

PR #436, opened 2023-07-08
feat: `dbQuoteLiteral()` correctly quotes 64-bit integers from the bit64 package (of class `"integer64"`)

### LINKED ISSUE #435: Bigints in dbQuoteLiteral
I noticed that `dbQuoteLiteral()` casts integer64 values as floats:
``` r
library("DBI")
conn <- dbConnect(RPostgres::Postgres())

(x <- bit64::as.integer64(1))
#> integer64
#> [1] 1
dbQuoteLiteral(conn, x)
#> <SQL> 1::float8
```
I'd expect `<SQL> 1::int8` instead. Would you consider a PR to support bigints here?

karawoo:
Fixes #435 by adding a branch of logic to `dbQuoteLiteral_PqConnection` that checks if the value is integer64, and if so appending `::int8`.

### DIFF
--- a/R/dbQuoteLiteral_PqConnection.R
+++ b/R/dbQuoteLiteral_PqConnection.R
@@ -22,6 +22,10 @@ dbQuoteLiteral_PqConnection <- function(conn, x, ...) {
     ret <- paste0("'", as.character(hms::as_hms(x)), "'")
     ret[is.na(x)] <- "NULL"
     SQL(paste0(ret, "::interval"), names = names(ret))
+  } else if (inherits(x, "integer64")) {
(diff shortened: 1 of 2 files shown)


krlmlr:
Thanks! I wonder why not use `bit64::as.integer64(1)` in the test, but this is a detail.

karawoo:
Sorry, for some reason I thought RPostgres didn't have a direct dependency on bit64, but I see that it actually does. I can update if you want.

krlmlr:
Thanks, there's no need -- it's easily done when we touch the code the next time.

PR #442, opened 2023-10-08
chore: Windows: update to libpq-15.3

jeroen:
This also adds support for ARM, which CRAN is going to support soon.


jeroen:
I think this is good to merge. The Windows compiler warning is triggered by your vendored boost, and is unrelated to this PR:

```r
Found the following significant warnings:
  vendor/boost/intrusive/detail/slist_node.hpp:56:16: warning: pointer used after 'void operator delete(void*, std::size_t)' [-Wuse-after-free]
  vendor/boost/intrusive/detail/hook_traits.hpp:77:10: warning: pointer used after 'void operator delete(void*, std::size_t)' [-Wuse-after-free]
```

krlmlr:
Good catch regarding this warning. Is this a g++12 bug? I could update the vendored headers to see if it goes away, but I wonder if an Rtools update is pending that will upgrade GCC.

krlmlr:
Thanks. It's great to have these warnings appear in CI/CD too 😊 . The same warnings are seen on CRAN, so perhaps not worth holding our breath.

jeroen:
Yes I think it is a bug in gcc-12. It is fixed in gcc-13. I can suggest a hack if you like.

krlmlr:
A hack would be great, thanks!

jeroen:
OK here you go: https://github.com/r-dbi/RPostgres/pull/443

PR #443, opened 2023-10-09
chore: Suppress warning in gcc-12

jeroen:
There is bug in gcc-12 that leads to the warning:

```
warning: pointer used after 'void operator delete(void*, std::size_t)' [-Wuse-after-free]
```

The bug is fixed in gcc-13. So we suppress it.


krlmlr:
Thanks!

PR #447, opened 2023-11-08
chore: Move libpq underneath `src/` on Windows

krlmlr:
Fixes warnings on Windows, I wonder why I don't see them on CRAN.


jeroen:
> Fixes warnings on Windows, I wonder why I don't see them on CRAN.

It only happens when building the *source package* on Windows. For the CRAN version, the source package was built by you on linux/mac, and only the binaries get built on Windows.

I prefer this solution: https://github.com/r-dbi/RPostgres/pull/449

PR #460, opened 2024-04-01
fix: Fix `dbQuoteIdentifier()` for `Id()` objects to no longer rely on names

krlmlr:


iangow:
> Fix `dbQuoteIdentifier()` for `Id()` objects to no longer rely on names.

More like "no longer *even uses* names" … this was a "breaking change" for me, as I was using `Id(table = "y", schema = "x")` in places [here](https://iangow.github.io/far_book/parquet-wrds.html). (I hadn't realized it hadn't been implemented already, so I thought I was safe when sending in my book.) All good now, it seems.

krlmlr:
Ian, I'm getting:

``` r
DBI::Id(table = "y", schema = "x")
#> <Id> "x"."y"
```

<sup>Created on 2024-06-08 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Do you expect a different result?

Anyway, thanks for the feedback, apologies for the inconvenience.

iangow:
I thought I was 

> Ian, I'm getting:
> 
> ```r
> DBI::Id(table = "y", schema = "x")
> #> <Id> "x"."y"
> ```
> 
> Created on 2024-06-08 with [reprex v2.1.0](https://reprex.tidyverse.org)
> 
> Do you expect a different result?
> 
> Anyway, thanks for the feedback, apologies for the inconvenience.

I thought I was seeing `Id(table = "y", schema = "x")` generating 'table "y"."x" not found' errors or similar, but now I can't remember precisely where I saw them and I can't reproduce them. I have reverted my code to `Id(table = "y", schema = "x")` and I'm not seeing any issues; (`<Id> "x"."y"`) is exactly what I want.

PR #465, opened 2024-04-19
Bump preferred libpq version on MacOS to 15

### LINKED ISSUE #441: Something wrong with the 1.4.5 Mac binary on CRAN with Redshift
I discovered an odd issue, which I thought I'd report and can help troubleshoot if needed. The tl;dr is that I cannot connect to one Redshift server with the Mac binary from CRAN, but can if I compile from source.

The current Mac binary of `RPostgres` (1.4.5) seems to have an issue with some, but not all, Redshift servers. The command I'm running is
```
DBI::dbConnect(RPostgres::Redshift(), ...) # same result with RPostgres::Postgres()
```
and the error I get is:

```
connection to server at "<host>" (<ip>), port 5439 failed: server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

The error message is the same also when I don't provide a valid username/password, so it seems to happen before authentication.

The weird thing is that I get this error with only one Redshift server (our production one) and not with another (our staging) – even though they're both running the same version. Also, I can connect to both with `psql` from my Mac, and can connect to both with `RPostgres` from within a docker container running on my Mac. This made me suspect it is somehow related to the Mac version of `RPostgres`. And in fact, if instead of installing the binary from CRAN I compile it from source with (I have `libpq` installed with Homebrew):
```
install.packages("RPostgres", type = "source")
```
The problem goes away and I can connect to both servers.

I don't know if the solution is to simply recompile the Mac binary version that is available on CRAN, and that will fix it. If not, I can help troubleshoot on my end, in case that helps. The fact that this seems to affect some, but not all, Redshift servers can understandably make this hard the reproduce. Is there something I can do to see debugging messages from `RPostgres`/`DBI`?

My `sessionInfo()`:
```
R version 4.2.3 (2023-03-15)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Big Sur ... 10.16

Matrix products: default
BLAS:   /Library/Frameworks/R.framework/Versions/4.2/Resources/lib/libRblas.0.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.2/Resources/lib/libRlapack.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

loaded via a namespace (and not attached):
 [1] bit_4.0.5       compiler_4.2.3  cli_3.6.1       hms_1.1.3      
 [5] DBI_1.1.3       Rcpp_1.0.11     bit64_4.0.5     vctrs_0.6.3    
 [9] blob_1.2.4      lifecycle_1.0.3 pkgconfig_2.0.3 rlang_1.1.1    
[13] RPostgres_1.4.5
```

jeroen:
Fall back on autobrew if the local macos libpq is out of date. Fixes #441.

### DIFF
--- a/configure
+++ b/configure
@@ -19,7 +19,7 @@ PKG_LIBS="-lpq"
 # Extra checks on MacOS for SSL support in libpq
 # command -v is probably fine: https://stackoverflow.com/a/677212/946850
 if [ "`uname`" = Darwin ] && [ "`command -v pkg-config`" ]; then
-  if pkg-config --atleast-version=12 libpq; then
+  if pkg-config --atleast-version=15 libpq; then
     case "`pkg-config --libs --static libpq`" in
     *crypto*)
       echo "Local libpq has SSL support"


krlmlr:
Thanks!

PR #472, opened 2024-10-19
feat: Importing large objects from client side

### LINKED ISSUE #376: How to write R object to Postgres large object
Cross-posted with https://stackoverflow.com/questions/70271590/read-write-postgres-large-objects-with-dbi-rpostgres

I'm trying to use the large object feature from Postgres https://www.postgresql.org/docs/10/largeobjects.html but I have a hard time wrapping my head around converting R object and writing them to the db.

Here is what I have tried so far: 

```bash
# Getting the db
docker run --rm --name some-postgres -e POSTGRES_PASSWORD=mysecretpassword -d -p 5433:5432 postgres
```

```r
library(DBI)
con <- dbConnect(
  RPostgres::Postgres(),
  dbname = "postgres",
  host = "localhost",
  port = 5433,
  user = "postgres",
  password = "mysecretpassword"
)
```

Creation works : 

```r
> dbGetQuery(con, "SELECT lo_create(1234);")
  lo_create
1      1234
```

But then I have a hard time figuring out how to write an R object to this large object. 
For example, how would I write `mtcars` as a large object in Postgres using `{DBI}` and `{RPostgres}`? 

And then, how do I read it back again in R?

toppyy:
Closes #376. Introduces a postgres-specific function `postgresImportLargeObject` for importing large objects from client side using libpq's `lo_import_with_oid`


R/PqConnection.R
```diff
@@ -135,3 +135,38 @@ postgresWaitForNotify <- function(conn, timeout = 1) {
...
+#' })
+postgresImportLargeObject <- function(conn, filepath = NULL, oid = 0) {
+
+  if (!postgresIsTransacting(conn)) {
+    stopc("Cannot import a large object outside of a transaction")
+  }
+
+  if (is.null(filepath)) stopc("'filepath' cannot be NULL")
+  if (oid <  0)     stopc("'oid' cannot be negative")
+  if (is.null(oid) | is.na(oid)) stopc("'oid' cannot be NULL/NA")
```
krlmlr, line 165:
```suggestion
  if (is.null(oid) || is.na(oid)) stopc("'oid' cannot be NULL/NA")
```

Or perhaps this could be two checks?

R/PqConnection.R
```diff
@@ -135,3 +135,38 @@ postgresWaitForNotify <- function(conn, timeout = 1) {
...
+  if (file.access(filepath,4) == -1) stopc(paste0("Unable to read from filepath '",filepath,"'"))
```
krlmlr, line 166: What will the error look like if we let libpq do the check?

> Please note that it is not a good idea to use this function to test before trying to open a file. On a multi-tasking system, it is possible that the accessibility of a file will change between the time you call `file.access()` and the time you try to open the file.

krlmlr (commented):
Thanks, nice! Checks are still failing: https://github.com/r-dbi/RPostgres/actions/runs/11416140787/job/31767025307?pr=472#step:16:263

R/PqConnection.R
toppyy: libpq returns an InvalidOid (a zero). Letting libpq do the check has the downside that we cannot inform the user of the specific problem as libpq's  `lo_import_with_oid()` returns a zero if the 1) file does not exists or 2) the assigned oid is already in use

toppyy: Thanks! I was unaware of `PQerrorMessage()` - that solves the problem! With the changes in the latest commit, when importing to an existing oid we get: 
> Error: ERROR:  duplicate key value violates unique constraint "pg_largeobject_metadata_oid_index"
DETAIL:  Key (oid)=(999) already exists.

(It's a bit clunky as the term "error" is repeated, but informative nonetheless.)

pqlib auto-assigns the oid if the argument 'oid' is zero. From the [docs](https://www.postgresql.org/docs/current/lo-interfaces.html#LO-IMPORT): 
> If lobjId is InvalidOid (zero) then lo_import_with_oid assigns an unused OID (this is the same behavior as lo_import).

I also changed the return- and argument types to 'Oid' from int as oid's are implemented as [unsigned integers](https://www.postgresql.org/docs/current/datatype-oid.html).

krlmlr:
Thanks!

PR #486, opened 2025-02-07
Windows: use libpq from Rtools if available

jeroen:
To comply with CRAN policies.


krlmlr:
Thanks. Do you think it's worth enabling checks for older versions of Windows in this repository? What about RMariaDB?

jeroen:
Yes I think that would be a good idea. I could not figure out how to enable it.

jeroen:
Could you show how to enable CI with older versions of Windows in your workflows?

krlmlr:
I might have responded elsewhere, just for completeness: We now check R 4.2 and later on Windows. The logic lives in https://github.com/r-dbi/RPostgres/blob/main/.github/versions-matrix.R .

PR #581, opened 2026-06-09
feat: expose PostgreSQL SQLSTATE and diagnostic fields on errors 

### LINKED ISSUE #579: Expose SQLSTATE (and other diagnostic fields) on error conditions
## Summary

When a query triggers a server-side error, RPostgres surfaces only the libpq
*message string* on the R condition. The SQLSTATE code and the other diagnostic
fields libpq exposes through `PQresultErrorField()` are discarded before the
error reaches R. As a result there is no reliable, locale-independent way to
classify an error programmatically.

It would be very useful if RPostgres carried at least the SQLSTATE onto the
condition.

## Why this matters

We define server-side validation in PL/pgSQL triggers and raise errors with
stable, explicit error codes:

```sql
RAISE EXCEPTION 'konto % ist nicht bebuchbar.', NEW.konto
  USING ERRCODE = 'FS002',
        HINT    = 'Setzen Sie bebuchbar = TRUE ...';
```

On the R side we want to branch on the *code* to drive the UI, precisely because
the message is user-facing, localized (German here), and may be reworded. The
code is the stable contract; the message is not.

Because only the message string reaches R, we are forced to recover the
classification by regex-matching the (German) message text:

```r
sqlstate <- if (grepl("ist nicht bebuchbar", msg, fixed = TRUE)) {
  "FS002"
} else if (grepl("nicht gültig", msg, fixed = TRUE)) {
  "FS001"
} else {
  NA_character_
}
```

This is brittle: it breaks if a message is reworded, it is locale-dependent, and
it cannot distinguish our raised codes from coincidentally similar text. The
`ERRCODE` we set on the server is the correct key to branch on — we just can't
read it back.

## Reproducible example

```r
library(DBI)
con <- dbConnect(RPostgres::Postgres())

err <- tryCatch(
  dbExecute(con, "DO $$ BEGIN RAISE EXCEPTION 'boom' USING ERRCODE = 'FS002'; END $$;"),
  error = function(e) e
)

conditionMessage(err)
#> [1] "Failed to fetch row : ERROR:  boom\nCONTEXT:  PL/pgSQL function inline_code_block line 1 at RAISE\n"

str(unclass(err))
#> List of 2
#>  $ message: chr "Failed to fetch row : ERROR:  boom\nCONTEXT:  PL/pgSQL ..."
#>  $ call   : NULL

class(err)
#> [1] "simpleError" "error"       "condition"
```

It is a plain `simpleError`: just `message` and `call`, no SQLSTATE field and no
SQLSTATE-derived class. The `'FS002'` code is gone. There is also no
`dbGetException()` fallback:

```r
existsMethod("dbGetException", "PqConnection")
#> [1] FALSE
```

## Where it's dropped

Every server-side error funnels through one helper, which keeps only
`PQerrorMessage()`:

```cpp
// src/DbConnection.cpp
void DbConnection::conn_stop(PGconn* conn, const char* msg) {
  cpp11::stop(std::string(msg) + " : " + PQerrorMessage(conn));
}
```

Result-level errors take the same path (`PqResultImpl::conn_stop` →
`DbConnection::conn_stop`), and `cpp11::stop()` raises a plain `simpleError` with
no extra fields. `PQresultErrorField` / `PG_DIAG_SQLSTATE` do not appear anywhere
in the source, so the SQLSTATE that libpq has on hand is never read.

## Desired behavior

Any of the following would solve it, in rough order of preference:

1. Attach the SQLSTATE as a field on a classed condition (e.g. `err$sqlstate`),
   and optionally give the condition an extra class such as
   `RPostgres_error_<sqlstate>` so handlers can dispatch directly:
   `tryCatch(..., RPostgres_error_FS002 = ...)`.
2. Additionally attach the other `PQresultErrorField()` diagnostics (detail,
   hint, column, constraint, table, schema, context) as fields, so callers don't
   have to parse them out of the message string.
3. At minimum, provide a `dbGetException()`-style accessor returning the SQLSTATE
   of the last error on the connection.

libpq already exposes everything needed via `PQresultErrorField(res, field)` —
`PG_DIAG_SQLSTATE`, `PG_DIAG_MESSAGE_DETAIL`, `PG_DIAG_MESSAGE_HINT`,
`PG_DIAG_COLUMN_NAME`, `PG_DIAG_TABLE_NAME`, `PG_DIAG_CONSTRAINT_NAME`, etc.
These would need to be read where the error result is turned into an R error,
and carried onto the condition rather than collapsed into a single string.

schochastics:
fix #579

When a server-side error occurs, RPostgres currently surfaces only libpq's error message string and discards the SQLSTATE code and the other diagnostic fields, which forces users to regex-match localized message text to classify errors. 

This PR captures those fields from the failing `PGresult` and raises a classed R condition that carries them: the condition gains the class vector `c("RPostgres_error_<SQLSTATE>", "RPostgres_error", "error", "condition")` and fields `sqlstate`, `hint`, `detail`, `context`, `table`, `column`, `datatype`, `constraint`, and `schema` (each `NULL` when libpq doesn't supply it). 

Users can now dispatch on a specific code (`tryCatch(..., RPostgres_error_FS002 = ...)`) or read `cnd$sqlstate` directly, including for custom codes set via `RAISE EXCEPTION ... USING ERRCODE`. The original error message is preserved unchanged, and all errors—including connection-level ones with no result—gain the `RPostgres_error` parent class for uniform handling. 

The condition is built in a small internal R helper invoked from C++; error sites that have a `PGresult` (`step_run`, `prepare`, and the COPY completion check) now route through it, while result-less paths keep their existing behavior aside from the new parent class. Includes tests covering undefined-table (42P01), syntax (42601), custom `ERRCODE` (FS002), and unique-violation (23505) cases.


krlmlr (commented):
Thanks.

- Could we also throw rlang errors with this approach?
- I'd like to include snapshot tests, needs `local_edition(3)` perhaps
- The approach differs from what duckdb is doing. I don't remember how I landed on the duckdb approach, perhaps just because I wanted rlang errors everywhere. I'm curious how and why this even works: R -> C++ -> R -> throwing and the resulting object survives. What does the captured call stack look like?

krlmlr:
Needs pkgdown config: https://github.com/r-dbi/RPostgres/actions/runs/27250549490/job/80473863145?pr=581#step:19:606 .

schochastics:
Ah sorry, missed the part about the rlang errors. I have another look

krlmlr:
Red CI/CD.

schochastics:
Claude says:
The remaining red checks are pre-existing CI issues, not from this PR — everything this PR touches is green

The two failing job types are unrelated to these changes:

1. **All `windows-latest` jobs** die in `after-install`:
   ```
   ankane/setup-postgres@v1/index.js:136
       throw `Postgres version not supported on Windows: ${postgresVersion}`;
   Postgres version not supported on Windows: 17
   ```
   `after-install/action.yml` defaults to `pg-version 17`, which `ankane/setup-postgres@v1` can't install on Windows, so Postgres never starts and the job aborts before the build/tests.

2. **`ubuntu-24.04 (devel)`** errors on NOTEs (devel uses `error-on: "note"`), and there are two long-standing ones: `Non-standard file/directory found at top level: 'configure.log'` (left by the package's own `configure` script) and `checking compilation flags in Makevars`.

krlmlr:
Yes, of course. Apologies, thanks for checking.

